PR #7, opened 2019-11-20
StepInput placeholder without key should resolve to $

andydouglas-exs:
Previously using the following syntax: `"ParamI": StateTwo.output()` failed with 

> ValueError: State 'StateThree' is using an illegal placeholder for the 'ParamI' parameter.

I've added a simple fix and extended the existing test cases so that `StepInput` placeholders behave similarly to `ExecutionInput`

AD

### DIFF
--- a/src/stepfunctions/inputs/placeholders.py
+++ b/src/stepfunctions/inputs/placeholders.py
@@ -125,6 +125,17 @@ class Placeholder(object):
         Returns:
             bool: `True` if placeholder variable was found in the collection. `False`, otherwise.
         """
+        for k, v in self.store.items():
+            if placeholder == v:
+                return True
+            elif v.contains(placeholder):
+                return True
+        return False
+
+    def __contains__(self, placeholder):
+        """
+            Containment check operator for placeholder variables.
+        """
         return self.contains(placeholder)
 
     def validate(self, input):
--- a/src/stepfunctions/inputs/placeholders.py
+++ b/src/stepfunctions/inputs/placeholders.py
@@ -125,7 +125,7 @@ class Placeholder(object):
         Returns:
             bool: `True` if placeholder variable was found in the collection. `False`, otherwise.
         """
-        return v.contains(placeholder)
+        return self.contains(placeholder)
 
     def validate(self, input):
         """
--- a/src/stepfunctions/inputs/placeholders.py
+++ b/src/stepfunctions/inputs/placeholders.py
@@ -125,12 +125,7 @@ class Placeholder(object):
         Returns:
             bool: `True` if placeholder variable was found in the collection. `False`, otherwise.
         """
-        for k, v in self.store.items():
-            if placeholder == v:
-                return True
-            elif v.contains(placeholder):
-                return True
-        return False
+        return v.contains(placeholder)
 
     def validate(self, input):
         """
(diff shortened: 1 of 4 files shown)


src/stepfunctions/steps/states.py
```diff
@@ -701,7 +701,7 @@ def visit(self, state):
 
     def _validate_next_step_params(self, params, step_output):
         for k, v in params.items():
-            if isinstance(v, StepInput) and not step_output.contains(v):
+            if isinstance(v, StepInput) and v.name and not step_output.contains(v):
```
shunjd: Hi Andrew,

Thank you very much for the fix, we recommend the following updates to make the code more readable.

```python
            if isinstance(v, StepInput):
                if v is not step_output and v not in step_output:
                    return False, k
```

This requires a change in the [placeholders.py](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/master/src/stepfunctions/inputs/placeholders.py#L134) with the following modifications.

```python
    def __contains__(self, placeholder):
        return self.contains(placeholder)
```

Could you submit another revision, we will be very happy to merge it.

Best,
Shunjia

andydouglas-exs: Thanks

Re: the first change. Minor - this does change the logic slightly i.e. it will no longer potentially enter the elif if `step_output.contains(v)` is true. This is probably fine and the tests pass cleanly.

The second suggested change causes infinite recursion in a number of tests
```
src/stepfunctions/inputs/placeholders.py:128: in contains
    return self.contains(placeholder)
src/stepfunctions/inputs/placeholders.py:128: in contains
    return self.contains(placeholder)
E   RecursionError: maximum recursion depth exceeded
!!! Recursion detected (same locals & position)
```
Please let me know how you want to want to proceed.

Andy

shunjd: For the second part, we have tested it locally here, it worked. Could you make a commit so that we can review the change?

Shunjia

src/stepfunctions/steps/states.py
```diff
@@ -701,8 +701,9 @@ def visit(self, state):
 
     def _validate_next_step_params(self, params, step_output):
         for k, v in params.items():
-            if isinstance(v, StepInput) and v.name and not step_output.contains(v):
-                return False, k
+            if isinstance(v, StepInput):
+                if v.name and not step_output.contains(v):
```
shunjd: The proper fix should not be checking the `v.name` but to verify if `v` is the `step_output`.

For instance, if this is the input.
```json
{
    "ParamI": StepTwo.output()
}
```

This will make `not step_output.contains(v)` to be False because `StepTwo.output().contains(StepTwo.output()) == False`. Because we don't want to fail the validation, we only need to check if the output itself is being passed.

src/stepfunctions/inputs/placeholders.py
```diff
@@ -125,12 +125,7 @@ def contains(self, placeholder):
         Returns:
             bool: `True` if placeholder variable was found in the collection. `False`, otherwise.
         """
-        for k, v in self.store.items():
-            if placeholder == v:
-                return True
-            elif v.contains(placeholder):
-                return True
-        return False
+        return self.contains(placeholder)
```
shunjd: You should keep the old logic here, and add a new method called `__contains__` so that we can make `v not in step_output` work.

```python
    def __contains__(self, placeholder):
        return self.contains(placeholder)
```

andydouglas-exs: Apologies, I missed the dunder I've reinstated and added a new method

src/stepfunctions/steps/states.py
andydouglas-exs: Not quite sure what you mean here, can you explain?

PR #11, opened 2019-12-17
Add service integrations for SQS, SNS and DynamoDB

shunjd:
*Issue #, if available:* N/A
*Description of changes:* Adding SQS, SNS and DynamoDB support.

### DIFF
--- a/src/stepfunctions/steps/service.py
+++ b/src/stepfunctions/steps/service.py
@@ -122,7 +122,7 @@ class SnsPublishStep(Task):
         super(SnsPublishStep, self).__init__(state_id, **kwargs)
 
 
-class SqsSendMessage(Task):
+class SqsSendMessageStep(Task):
 
     """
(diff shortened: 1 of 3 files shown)


src/stepfunctions/steps/service.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        if wait_for_callback:
+            kwargs[Field.Resource.value] = 'arn:aws:states:::sns:publish.waitForTaskToken'
+        else:
+            kwargs[Field.Resource.value] = 'arn:aws:states:::sns:publish'
+        
+        super(SnsPublishStep, self).__init__(state_id, **kwargs)
+
+
+class SqsSendMessage(Task):
```
vaib-amz: Do you think it makes sense to rename this class to `SqsSendMessageStep` with the `Step` suffix for consistency? That seems to be the pattern being followed here.

shunjd: Good catch, I've changed it in the next revision.

Issue #22, opened 2020-01-31
feature: Add support for passing ExperimentConfig to training and transform steps


brightsparc:
When creating a training or transform job an orphaned "TrailComponent" will be created.

The `TrainingStep` and `TransformStep` constructors could be extended to support passing in this ExperimentConfig dictionary eg:

```
    {
                'ExperimentName': 'pca_experiment',
                'TrialName': 'pca_trial',
                'TrialComponentDisplayName': 'Transform'                
    }
```

brightsparc:
Closing because merged.

PR #23, opened 2020-01-31
feature: adding support for ExperimentConfig in training and transform

brightsparc:
*Issue #22:*

*Description of changes:*

I've added additional optional arguments to the `TrainingStep` and `TransformStep` classes that will set the dictionary as properties that will then be passed through to the steps for execution.


brightsparc:
I'm getting error `Unable to locate credentials` locally when trying to run unit tests, not sure if this is related to mock boto not correctly registering.  I get the same error on previous commit from @shunjd

PR #37, opened 2020-03-31
fix: add tags for sagemaker

gercograndia:
*Issue #, Most of the sagemaker steps did not pass along tags, which made it unusable for us as we require these tags for authorisation as well as cost allocation purposes

*Description of changes:*
For most of the Sagemaker steps I have added an optional tags parameter to the __init__, and included that in the Parameters dictionairy.

Also added the tags to the relevant unit tests.


vaib-amz:
Thanks for creating this pull request @gercograndia . Looks like all approvals have been met. Merging now.

PR #47, opened 2020-05-07
Doc: fix definition parameter to State or Chain

tuliocasagrande:
*Description of changes:*

This PR fixes the documentation for the definition parameter which expects a State or Chain and not a string.

Also add `Args:` to update() to correctly render in the doc page: https://aws-step-functions-data-science-sdk.readthedocs.io/en/latest/workflow.html#stepfunctions.workflow.Workflow.update.


vaib-amz:
Hey, thanks for the fix!

Vaib

shunjd:
Thanks a lot for the fix. @tuliocasagrande

Issue #45, opened 2020-05-06
Cannot output null value in ResultPath


shunjd:
We cannot output the following JSON definition even if result_path is None.

```
{
  "Type": "Parallel",
  "ResultPath": null
}
```

If `result_path` is specified explicitly, user should expect to have it in the final statemachine definition.

```
from stepfunctions import steps

p = steps.Parallel('test', result_path=None)
assert 'ResultPath' in p.to_dict()
```

tuliocasagrande:
Hey @shunjd, I hope you're doing great man.

I faced the same behavior yesterday with a `steps.Task`, so I submitted #48 that I think fixes this error for any state type.

Let me know what you think.
Thanks!

PR #48, opened 2020-05-08
Fix null value in ResultPath

tuliocasagrande:
*Issue #, if available:*
aws/aws-step-functions-data-science-sdk-python#45

*Description of changes:*
This PR fixes the output null value in ResultPath to behave as described here: https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html#input-output-resultpath-null

The argument `result_path=None` in Python is going to be converted to `"ResultPath": null` in JSON.

I tested with Task, but it could be tested with any state type as I believe `"ResultPath": null` is a behavior available for any type.

Thank you for reviewing this!

### DIFF
--- a/src/stepfunctions/steps/states.py
+++ b/src/stepfunctions/steps/states.py
@@ -68,7 +68,7 @@ class Block(object):
         result = {}
         # Common fields
         for k, v in self.fields.items():
-            if v is not None:
+            if v is not None or k == 'result_path':
                 k = to_pascalcase(k)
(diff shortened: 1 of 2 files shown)


tuliocasagrande:
Sure, no problem.
I didn't know `"InputPath": null` and `"OutpuPath": null` were also valid options.

tuliocasagrande:
@shunjd I pushed a new commit with `InputPath` and `OutputPath`. Let me know if the tests are OK or if it's better to make one test for each parameter.

Thanks again!

Issue #50, opened 2020-05-22
LambdaStep using one of two possible options to call Lambda


tvaroska:
Step Function can call Lamba in two ways - with lambda name in Parameters -> FunctionName or with ARN in Resource. Second one does simplify data moves between functions and allows links from the console to CloudWatch logs etc.

Both ways are described in https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html

shunjd:
Hi @tvaroska, there is a simple way of doing that.

```python
from stepfunctions.steps.states import Task

lambda_task = Task(resource='your_lambda_arn')
```

tvaroska:
agree, it is possible.
what I would like to improve is clarity of API. using Task for prefered way to call Lambda and LambdaTask for the secondary is not exactly clear

shunjd:
It is a design decision we had made to promote the usage of the new [Lambda Invoke](https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html), this pattern is more consistent with other service integrations. The old way of using Lambda by putting the LambdaArn into the `Resource` field has its own limitation, it does not allow the customer to specify the `Parameters` field.

shunjd:
As mentioned in https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/50, we would not want to update the LambdaStep for this change, I will close the PR.

PR #51, opened 2020-05-22
#50 Modify LambdaStep to use Resource as Lambda ARN

tvaroska:
Issue #50  

Modification of LambdaStep to accept parameter lambda_arn. If parameter exists it will use it to create Lambda step in format:
      "Step_id": {  
         "Type":"Task",
         "Resource":"lambd_arn"
      }

as described in step function documentation - https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html

Unit tests passed.


shunjd:
As mentioned in #50, we would not want to update the LambdaStep for this change, I will close the PR.

Issue #82, opened 2020-09-01
Problem with environment variables in modelstep


RasmusToivanen:
Hi,
Please help.
I am having some challenges with stepfunctions + sagemaker APIs.
It seems that I should be able to pass environmental variables to endpoint using env(dict[str,str])  in the sagemaker.Model constructor in the steps.Modelstep. 
However I am not able to get those env variables visible in the endpoint.
I am able to get the env variables working with boto3 but I don't want to use that as I prefer Sagemaker API with step functions.

So how to pass env variables correctly to stepfunctions to that those are visible?

Running:
stepfunctions=1.0.0.9
sagemaker=1.55.4

![image](https://user-images.githubusercontent.com/69349765/91888881-a2924a80-ec95-11ea-904c-8d9fd724eccf.png)
However in container once I start it (printing os.environ all keys and values)
![image](https://user-images.githubusercontent.com/69349765/91888981-c5246380-ec95-11ea-88a4-85996ba9b00a.png)

RasmusToivanen:
With boto3 I can pass values in the same way:
Definition:
![image](https://user-images.githubusercontent.com/69349765/91897359-d031c080-eca2-11ea-9ec4-339944c98439.png)
Same container printing os.environ
![image](https://user-images.githubusercontent.com/69349765/91897424-f0617f80-eca2-11ea-92e4-b8b7ae926216.png)

t6nn:
Looking at the source for [ModelStep](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/b9b0371698a9a9c154066eba442085d0e1dc722e/src/stepfunctions/steps/sagemaker.py#L213), it seems that the supplied environment is currently ignored. 

Not sure what the effort / consequences to fix this would be, but perhaps a temporary workaround could be to implement your own ModelStep class with environment support or to update the environment directly in the State? For example, something in this direction (haven't tried it myself):
`dev_model_step.fields['parameters']['PrimaryContainer']['Environment'] = { ... your environment here ...}`

RasmusToivanen:
Thans really much @t6nn,
I will try your recommended approaches and will come back to you once I see whether I get those working.

ca-nguyen:
Thank you for reporting this bug!
The fix will be included in the next release

PR #84, opened 2020-09-08
Made it so that underlying Model object makes use of environmental va…

R4ZZ3:
…riables which are then passed to endpoint

*Issue #, if available:*
#82

*Description of changes:*
I had problems with passing env variables to endpoints using Modelstep.
It turned out it wasn't passed to model parameters in the modelstep at all.
Added it so it will now be correctly passed. Tested in one of my projects and it seemed to work


ca-nguyen:
Thank you for contributing!

Closing this PR - Superseded by #160

PR #88, opened 2020-09-21
[Feature]: Support new ASL choice comparators

yoodan93:
Adding new choice state comparators that were added to the Amazon States Language specifications. https://states-language.net/

These include:

- StringMatches
- IsNull, IsPresent, IsString, IsNumeric, IsBoolean, IsTimestamp
- StringEqualsPath, StringLessThanPath, StringGreaterThanPath, StringLessThanEqualsPath, StringGreaterThanEqualsPath, NumericEqualsPath, NumericLessThanPath, NumericGreaterThanPath, NumericLessThanEqualsPath, NumericGreaterThanEqualsPath, BooleanEqualsPath, TimestampEqualsPath, TimestampLessThanPath, TimestampGreaterThanPath, TimestampLessThanEqualsPath, TimestampGreaterThanEqualsPath


tests/unit/test_choice_rule.py
```diff
@@ -61,6 +61,35 @@ def test_variable_value_must_be_consistent():
         with pytest.raises(ValueError):
             func('$.Variable', True)
 
+def test_path_function_value_must_be_consistent():
```
wong-a, line 64:
```suggestion
def test_path_operator_raises_error_when_value_is_not_a_path():
```

tests/unit/test_choice_rule.py
```diff
@@ -61,6 +61,35 @@ def test_variable_value_must_be_consistent():
...
+        'NumericEqualsPath',
+        'TimestampEqualsPath',
+        'BooleanEqualsPath'
+    }
+    for path_function in path_functions:
+        func = getattr(ChoiceRule, path_function)
+        with pytest.raises(ValueError):
+            func('$.Variable', 'string')
+
+def test_is_function_value_must_be_consistent():
```
wong-a, line 76:
```suggestion
def test_is_operator_raises_error_when_value_is_not_a_bool():
```

src/stepfunctions/steps/choice_rule.py
```diff
@@ -376,6 +592,120 @@ def TimestampGreaterThanEquals(cls, variable, value):
...
+        return Rule(variable, 'TimestampGreaterThanEqualsPath', value)
+
+    @classmethod
+    def IsNull(cls, variable, value):
+        """
+        Creates a rule with the `IsNull` operator.
+
+        Args:
+            variable (str): Path to the variable to compare.
+            value (bool): Constant value to compare `variable` against.
```
wong-a, line 616: For the type comparisons, maybe the documentation for `value` can be more descriptive:

- IsNull: "Whether the value at `variable` is equal to the JSON literal `null` or not.
- IsPresent: "Whether a field at `variable` exists in the input or not.
- IsNumeric: "Whether the value at `variable` is a string or not.
- etc.

tests/unit/test_choice_rule.py
```diff
@@ -61,6 +61,35 @@ def test_variable_value_must_be_consistent():
...
+    }
+
+    for is_function in is_functions:
+        func = getattr(ChoiceRule, is_function)
+        with pytest.raises(ValueError):
+            func('$.Variable', 'string')
+        with pytest.raises(ValueError):
+            func('$.Variable', 101)
+
 def test_rule_serialization():
```
wong-a, line 93: Can we add similar tests like this for the other choice rules to make sure they are serialized to the correct ASL? Since the rules that aren't boolean expressions more or less have the same pattern, it should be easy to iterate over them.

wong-a (changes requested):

vaib-amz (dismissed):

rwwardh (approved):
Adam mentioned some naming changes. Once done it should be good to go.

src/stepfunctions/steps/choice_rule.py
yoodan93: fixed in subsequent commit

tests/unit/test_choice_rule.py
yoodan93: added test for all choice rules serialization

src/stepfunctions/steps/choice_rule.py
```diff
@@ -376,6 +592,120 @@ def TimestampGreaterThanEquals(cls, variable, value):
...
+        return Rule(variable, 'IsBoolean', value)
+
+    @classmethod
+    def StringMatches(cls, variable, value):
+        """
+        Creates a rule with the `StringMatches` operator.
+
+        Args:
+            variable (str): Path to the variable to compare.
+            value (bool): Constant value to compare `variable` against.
```
wong-a, line 700: I missed this one earlier. The value for `StringMatches` is a `str`. Since this is one of the ambiguous operators, let's add a description of the format in the documentation. Customers shouldn't have to go the ASL spec to figure out how to use this SDK.

```suggestion
            value (str): A string pattern that may contain one or more `*` characters to compare the value at `variable` to. The comparison yields true if the variable matches the pattern, where `*` is a wildcard that matches zero or more characters.
```

Trying to think of good wording about escaping special characters with `\`. Not sure if there's anything that's handled differently in Python.

wong-a (changes requested):

src/stepfunctions/steps/choice_rule.py
```diff
@@ -376,6 +592,122 @@ def TimestampGreaterThanEquals(cls, variable, value):
...
+
+    @classmethod
+    def StringMatches(cls, variable, value):
+        """
+        Creates a rule with the `StringMatches` operator.
+
+        Args:
+            variable (str): Path to the variable to compare.
+            value (str): A string pattern that may contain one or more `*` characters to compare the value at `variable` to.
+            The `*` character can be escaped using two backslashes.
```
wong-a, line 701: Not sure if this docstring will render this correctly if the newlines aren't indented

wong-a (dismissed):

wong-a (approved):

vaib-amz (approved):

PR #118, opened 2021-02-10
feat: add a new buildspec configuration that runs both unit and integ tests

shivlaks:
motivation: rather than running integ and unit tests in separate CodeBuild projects and for PR builds,
consolidate them into a single project which will run the sequentially. This is mostly to simplify the
current infrastructure setup and make PRs less confusing as we will 1 set of build logs for every PR build.

The current PR build runs these separately and posts logs/comments for each separately, which is more tedious.

---

### DIFF
--- a/buildspec_pr.yaml
+++ b/buildspec_pr.yaml
@@ -3,9 +3,8 @@ version: 0.2
 phases:
   pre_build:
     commands:
-      - pip install setuptools wheel twine
+      - pip install setuptools wheel twine tox
       - python setup.py sdist
-      - pip install tox
   build:
     commands:
       - twine check dist/*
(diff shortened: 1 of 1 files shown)


buildspec_pr.yaml
```diff
@@ -0,0 +1,13 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - pip install setuptools wheel twine
+      - python setup.py sdist
+      - pip install tox
```
wong-a, line 8: Can this be in the first `pip` command?

shivlaks: good call, i'll give that a try, but am unsure whether there's motivation behind the current structure which I reused.

shivlaks: verified that it can. updated the PR

shivlaks:
closing in favour of using it in infrastructure code rather than adding it to the repo

PR #130, opened 2021-04-29
Update README.rst

yuan-bwn:
*Issue #, if available:*
Fix broken image link in Readme
*Description of changes:*
Commit 1:
Use unprocessed version of image under  `raw.githubusercontent.com` domain to successfully reveal image. For rst tiles [the size adjust option is not working](https://github.com/github/markup/issues/295) in github, but it will work if fetched to local. 

Commit 2:
Use relative path for image links. Adjust image size displayed on local env.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -157,16 +157,16 @@ Using this SDK you can create steps, chain them together to create a workflow,
 create that workflow in AWS Step Functions, and execute the workflow in the
 AWS cloud.
 
-.. image:: https://github.com/aws/aws-step-functions-data-science-sdk-python/raw/master/doc/images/create.png
-  :scale: 50 %
-  :alt: Create a workflow in AWS Step Functions
+.. image:: https://raw.githubusercontent.com/aws/aws-step-functions-data-science-sdk-python/master/doc/images/create.png
+    :width: 400
+    :alt: Create a workflow in AWS Step Functions
 
 Once you have created your workflow in AWS Step Functions, you can execute that
 workflow in Step Functions, in the AWS cloud.
 
-.. image:: https://github.com/aws/aws-step-functions-data-science-sdk-python/raw/master/doc/images/execute.png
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -157,16 +157,16 @@ Using this SDK you can create steps, chain them together to create a workflow,
 create that workflow in AWS Step Functions, and execute the workflow in the
 AWS cloud.
 
-.. image:: https://github.com/aws/aws-step-functions-data-science-sdk-python/raw/master/doc/images/create.png
-  :scale: 50 %
-  :alt: Create a workflow in AWS Step Functions
+.. image:: https://raw.githubusercontent.com/aws/aws-step-functions-data-science-sdk-python/master/doc/images/create.png
```
wong-a, line 160: Good catch for broken images, but the raw URL isn't the problem. We switched branches from `master` -> `main` to use more inclusive language. It's GitHub's new default branch name now.

Can you use a relative path instead? That way it will also work between branches.
```suggestion
.. image:: ./doc/images/create.png
```

wong-a (changes requested):

README.rst
```diff
@@ -157,16 +157,16 @@ Using this SDK you can create steps, chain them together to create a workflow,
...
+    :width: 400
```
wong-a, line 161: Maybe 600px instead? 400 looks a little narrow.
```suggestion
    :width: 600
```

README.rst
yuan-bwn: Yes I can use a relative path. Will update it soon

README.rst
yuan-bwn: The usual recommended pixel range is 400 - 800 so I think 600 makes sense. Will make the change

wong-a (approved):

ca-nguyen (approved):

Issue #120, opened 2021-02-23
partitions(aws-cn, us-gov) support


zxkane:
All resources of task are hard code partition `aws`. The partition info should be fetched from boto3 client.

zxkane:
We are using below workaround for reference,

```python
from stepfunctions.steps.fields import Field

attrs=getattr(step, ‘fields’)
attrs[Field.Resource.value]=‘arn:aws-cn:states:::glue:startJobRun.sync’
setattr(step, ‘fields’, attrs)
```

wong-a:
Noted. Thanks for raising this!

Setting `Field.Resource.value` for `Steps` such as [this one](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/master/src/stepfunctions/steps/service.py#L38) need to be updated.

hlmiao:
How test the fixed version, same issue as previous experienced in cn-north-1.
{
  "error": "States.Runtime",
  "cause": "An error occurred while executing the state 'Extract, Transform, Load' (entered at the event id #2). The resource belongs to a different partition from the running execution."
}

wong-a:
@hlmiao https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/131 hasn't been merged or released yet. If you want to test it out, you can pull the source from the PR's branch.

hlmiao:
Got it.

wong-a:
Released in v2.1.0: https://github.com/aws/aws-step-functions-data-science-sdk-python/releases/tag/v2.1.0

PR #131, opened 2021-05-05
fix: make arns of all task resources aws-partition aware

yuan-bwn:
*Issue #, if available:* #120
*Update:*
Resolve previous comments. 
Use boto3 patch mock for unit tests. 
Use f-string for string building.
Create arn builder for providing integration resource arn.
Create tests for arn builder.

*Test:*
All unit tests and integ tests passed.

*Description of changes:*
Support different aws partitions when SDK providing resource arns. Partitions will be determined by boto3 client session running region.
Available regions are based on SFN regional availability.
There are formatting changes for test_steps.py which done by IDE. They are valid changes and will not have functionality impact.

*Test:*
Created new unit under test_steps suite. All units tests passed.
All integration tests passed in both `us-east-1` and `cn-north-1` using test account.

### DIFF
--- a/src/stepfunctions/steps/compute.py
+++ b/src/stepfunctions/steps/compute.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import
 
 from stepfunctions.steps.states import Task
 from stepfunctions.steps.fields import Field
+from stepfunctions.steps.utils import get_aws_partition
 
 
 class LambdaStep(Task):
@@ -38,9 +39,9 @@ class LambdaStep(Task):
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         if wait_for_callback:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::lambda:invoke.waitForTaskToken'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::lambda:invoke.waitForTaskToken'
         else:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::lambda:invoke'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::lambda:invoke'
 
         super(LambdaStep, self).__init__(state_id, **kwargs)
 
@@ -67,9 +68,9 @@ class GlueStartJobRunStep(Task):
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         if wait_for_completion:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::glue:startJobRun.sync'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::glue:startJobRun.sync'
         else:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::glue:startJobRun'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::glue:startJobRun'
 
         super(GlueStartJobRunStep, self).__init__(state_id, **kwargs)
 
@@ -96,9 +97,9 @@ class BatchSubmitJobStep(Task):
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         if wait_for_completion:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::batch:submitJob.sync'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::batch:submitJob.sync'
         else:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::batch:submitJob'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::batch:submitJob'
 
         super(BatchSubmitJobStep, self).__init__(state_id, **kwargs)
 
@@ -125,8 +126,8 @@ class EcsRunTaskStep(Task):
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         if wait_for_completion:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::ecs:runTask.sync'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::ecs:runTask.sync'
         else:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::ecs:runTask'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::ecs:runTask'
 
         super(EcsRunTaskStep, self).__init__(state_id, **kwargs)
--- a/src/stepfunctions/steps/compute.py
+++ b/src/stepfunctions/steps/compute.py
@@ -12,10 +12,32 @@
 # permissions and limitations under the License.
 from __future__ import absolute_import
 
+from enum import Enum
 from stepfunctions.steps.states import Task
 from stepfunctions.steps.fields import Field
-from stepfunctions.steps.utils import resource_integration_arn_builder
-from stepfunctions.steps.integration_resources import IntegrationPattern, IntegrationServices, LambdaApi, GlueApi, BatchApi, EcsApi
+from stepfunctions.steps.utils import get_service_integration_arn
+from stepfunctions.steps.integration_resources import IntegrationPattern
+
+Lambda = "lambda"
+Glue = "glue"
+Ecs = "ecs"
+Batch = "batch"
+
+
+class LambdaApi(Enum):
+    Invoke = "invoke"
+
+
+class GlueApi(Enum):
+    StartJobRun = "startJobRun"
+
+
+class EcsApi(Enum):
+    RunTask = "runTask"
+
+
+class BatchApi(Enum):
+    SubmitJob = "submitJob"
 
 
 class LambdaStep(Task):
@@ -44,14 +66,16 @@ class LambdaStep(Task):
             """
             Example resource arn: arn:aws:states:::lambda:invoke.waitForTaskToken
             """
-            kwargs[Field.Resource.value] =  resource_integration_arn_builder(IntegrationServices.Lambda,
-                                                                             LambdaApi.Invoke,
-                                                                             IntegrationPattern.WaitForTaskToken)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Lambda,
+                                                                        LambdaApi.Invoke,
+                                                                        IntegrationPattern.WaitForTaskToken)
         else:
             """
             Example resource arn: arn:aws:states:::lambda:invoke
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.Lambda, LambdaApi.Invoke)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Lambda, LambdaApi.Invoke)
 
 
         super(LambdaStep, self).__init__(state_id, **kwargs)
@@ -82,15 +106,17 @@ class GlueStartJobRunStep(Task):
             """
             Example resource arn: arn:aws:states:::glue:startJobRun.sync
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.Glue,
-                                                                            GlueApi.StartJobRun,
-                                                                            IntegrationPattern.WaitForCompletion)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Glue,
+                                                                       GlueApi.StartJobRun,
+                                                                       IntegrationPattern.WaitForCompletion)
         else:
             """
             Example resource arn: arn:aws:states:::glue:startJobRun
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.Glue,
-                                                                            GlueApi.StartJobRun)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Glue,
+                                                                       GlueApi.StartJobRun)
 
         super(GlueStartJobRunStep, self).__init__(state_id, **kwargs)
 
@@ -120,15 +146,17 @@ class BatchSubmitJobStep(Task):
             """
             Example resource arn: arn:aws:states:::batch:submitJob.sync
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.Batch,
-                                                                            BatchApi.SubmitJob,
-                                                                            IntegrationPattern.WaitForCompletion)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Batch,
+                                                                       BatchApi.SubmitJob,
+                                                                       IntegrationPattern.WaitForCompletion)
         else:
             """
             Example resource arn: arn:aws:states:::batch:submitJob
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.Batch,
-                                                                            BatchApi.SubmitJob)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Batch,
+                                                                       BatchApi.SubmitJob)
 
         super(BatchSubmitJobStep, self).__init__(state_id, **kwargs)
 
@@ -158,14 +186,16 @@ class EcsRunTaskStep(Task):
             """
             Example resource arn: arn:aws:states:::ecs:runTask.sync
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.ECS,
-                                                                            EcsApi.RunTask,
-                                                                            IntegrationPattern.WaitForCompletion)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Ecs,
+                                                                       EcsApi.RunTask,
+                                                                       IntegrationPattern.WaitForCompletion)
         else:
             """
             Example resource arn: arn:aws:states:::ecs:runTask
             """
-            kwargs[Field.Resource.value] = resource_integration_arn_builder(IntegrationServices.ECS,
-                                                                            EcsApi.RunTask)
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Ecs,
+                                                                       EcsApi.RunTask)
 
         super(EcsRunTaskStep, self).__init__(state_id, **kwargs)
--- a/src/stepfunctions/steps/compute.py
+++ b/src/stepfunctions/steps/compute.py
@@ -14,7 +14,8 @@ from __future__ import absolute_import
 
 from stepfunctions.steps.states import Task
 from stepfunctions.steps.fields import Field
-from stepfunctions.steps.utils import get_aws_partition
+from stepfunctions.steps.utils import resource_integration_arn_builder
+from stepfunctions.steps.integration_resources import IntegrationPattern, IntegrationServices, LambdaApi, GlueApi, BatchApi, EcsApi
 
 
 class LambdaStep(Task):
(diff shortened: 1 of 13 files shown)


tests/unit/test_steps.py
```diff
@@ -13,24 +13,28 @@
 from __future__ import absolute_import
 
 import pytest
+import boto3
+import os
 
 from stepfunctions.exceptions import DuplicateStatesInChain
-from stepfunctions.steps import Pass, Succeed, Fail, Wait, Choice, ChoiceRule, Parallel, Map, Task, Retry, Catch, Chain
+from stepfunctions.steps import Pass, Succeed, Fail, Wait, Choice, ChoiceRule, Parallel, Map, Task, Retry, Catch, Chain, \
+    utils
```
ca-nguyen, line 21: Should we reorganize the imports so that each is on another line to make is more readable? It would also make it easier to spot the change next time a new import is added from stepfunctions.step.

Something like: 
```
from stepfunctions.steps import (
    Catch,
    Chain,
    Choice,
    ChoiceRule,
    Fail,
    Map,
    Parallel,
    Pass,
    Retry,
    Succeed,
    Task,
    Wait,
    utils,
)
```

src/stepfunctions/steps/utils.py
```diff
@@ -12,6 +12,32 @@
...
 def tags_dict_to_kv_list(tags_dict):
-    kv_list = [{"Key": k, "Value": v} for k,v in tags_dict.items()]
-    return kv_list
\ No newline at end of file
+    kv_list = [{"Key": k, "Value": v} for k, v in tags_dict.items()]
+    return kv_list
+
+
+# Obtain matching aws partition name based on region
+# Retrun "aws" as default if no region detected
```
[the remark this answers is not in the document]
yuan-bwn: Nice catch!  I will make the change in next commit

tests/unit/test_steps.py
yuan-bwn: I was following the legacy across this repo and it has many places import like this. As these imports have relatively short name, to me importing same line is a better choice. Make the list longer is definitely clearer but the looking will be clumsy.
But let me know if this is hard to read. If so I can maybe divide them into 2 or 3 lines so that we have a clearer reading as well as keeping the context less bulky.

src/stepfunctions/steps/compute.py
```diff
@@ -38,9 +39,9 @@ def __init__(self, state_id, wait_for_callback=False, **kwargs):
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         if wait_for_callback:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::lambda:invoke.waitForTaskToken'
+            kwargs[Field.Resource.value] = 'arn:' + get_aws_partition() + ':states:::lambda:invoke.waitForTaskToken'
```
wong-a, line 42: _Not necessarily blocking:_ I would like to at least consider class for ARN construction instead of building strings with repeated calls to `get_aws_partition()` inline everywhere. The AWS SDK in other languages have utilities for this, but python doesn't.

Something like:
```python
kwards[Field.Resource.value] = ARN(service='states', resource='lambda:invoke')
```

Or even scoping it to service integration ARNs:
```python
kwargs[Field.Resource.value] = ServiceIntegrationARN(service='lambda', api='invoke', integration_pattern=IntegrationPattern.WAIT_FOR_TASK_TOKEN)
```

wong-a (changes requested):

src/stepfunctions/steps/utils.py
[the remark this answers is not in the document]
yuan-bwn: Sure thing. Will make the fix in new commit.

src/stepfunctions/steps/utils.py
```diff
@@ -12,6 +12,32 @@
...
+
+# Obtain matching aws partition name based on region
+# Retrun "aws" as default if no region detected 
+def get_aws_partition():
+    partitions = boto3.session.Session().get_available_partitions()
+    cur_region = boto3.session.Session().region_name
+    cur_partition = "aws"
+
+    if cur_region is None:
+        logger.warning("No region detected for the session, will use default partition: aws")
```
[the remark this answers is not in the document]
yuan-bwn: Makes sense. Will do

src/stepfunctions/steps/utils.py
```diff
@@ -12,6 +12,32 @@
...
+        return cur_partition
+
+    for partition in partitions:
```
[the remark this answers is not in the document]
yuan-bwn: Thanks. Makes sense. I saw this year-old PR in botocore but forgot to link it here.
That PR never got merged in and the request for this is high. I will make another request in boto3 repo

tests/unit/test_steps.py
```diff
@@ -424,3 +445,30 @@ def test_default_paths_not_converted_to_null():
     assert '"ResultPath": null' not in task_state.to_json()
     assert '"InputPath": null' not in task_state.to_json()
     assert '"OutputPath": null' not in task_state.to_json()
+
+
+# Test if boto3 session can fetch correct aws partition info from test environment
+def test_util_get_aws_partition():
```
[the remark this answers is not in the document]
yuan-bwn: Sure thing. Will do

tests/unit/test_steps.py
```diff
@@ -424,3 +445,30 @@ def test_default_paths_not_converted_to_null():
...
+    aws_partition = "aws"
+    aws_cn_partition = "aws-cn"
```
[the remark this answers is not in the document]
yuan-bwn: I was thinking it makes the reading better if we mention what we are testing but they both will work I think.

tests/unit/test_steps.py
```diff
@@ -424,3 +445,30 @@ def test_default_paths_not_converted_to_null():
...
+    default_region = None
+
+    # Boto3 used either info from ~/.aws/config or AWS_DEFAULT_REGION in environment
+    # to determine current region. We will replace/create AWS_DEFAULT_REGION with regions in
+    # different aws partition to test that when regions are changed, correct partition info
+    # can be retrieved.
```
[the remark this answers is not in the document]
yuan-bwn: No I didn't. I came up this testing flow during my local setup and think this will satisfy the unit test need.

tests/unit/test_steps.py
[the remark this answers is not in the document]
wong-a: This seems pretty obvious to me. But this is just nitpicking

```python
assert utils.get_aws_partition() == 'aws'
assert utils.get_aws_partition() == 'aws-cn'
```

src/stepfunctions/steps/compute.py
yuan-bwn: This will definitely make the code cleaner. I will look into it.
If I can not finish this quickly I will prepare a PR specific for it later.

tests/unit/test_steps.py
```diff
@@ -424,3 +445,30 @@ def test_default_paths_not_converted_to_null():
...
+
+
+# Test if boto3 session can fetch correct aws partition info from test environment
+def test_util_get_aws_partition():
+    aws_partition = "aws"
+    aws_cn_partition = "aws-cn"
+    default_region = None
+
+    # Boto3 used either info from ~/.aws/config or AWS_DEFAULT_REGION in environment
+    # to determine current region. We will replace/create AWS_DEFAULT_REGION with regions in
```
shivlaks, line 457: couple questions here:

1. What will happen if your assertion fails after `cn-north-1` is set as the default region? will those changes persist for the next boto session because the test will bail or is there a guarantee that environment variables that are being touched will always be restored to a pristine state?
2. does this work with named profiles and other mechanisms that can be used to establish aws credentials where the `DEFAULT_REGION` is not set as an environment variable?

It's generally not preferable for unit tests to have any dependencies. Extraneous processes could be running or setting the region and this test could interfere and introduce surprising behaviour by modifying them.

i know the [`stubber`](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/stubber.html#) allows you to create a session for a client with a specific region, but i'm not sure what mocking the session looks like.

shivlaks (changes requested):

tests/unit/test_steps.py
yuan-bwn: This makes sense. So far I think there are definitely different style existing in this repo.

tests/unit/test_steps.py
yuan-bwn: 1. Good question. From my understanding this will continue to fail as the `cn-north-1` will be set as env var which will keep being effective till next run. To improve this specific issue I can save exceptions as non-fatal and assert at the very end of test so that changes got reversed before assertion.
2. Yes. I tested with both environment variable and ~/.aws/config file. Can't think of other configuration methods but please let me know if you have some in your mind.
3. I do agree the test should be on itself and should be mocked without dependency. So far I tried to mock boto3 client/session but have no luck. Will give stubber a try.

src/stepfunctions/steps/integration_resources.py
```diff
@@ -0,0 +1,89 @@
...
+from enum import Enum
+
+"""
+Enum classes for task integration resource arn builder
+"""
+
+
+class IntegrationPattern(Enum):
+    WaitForTaskToken = "waitForTaskToken"
+    WaitForCompletion = "sync"
```
wong-a, line 25: Let's model the default "Request Response" pattern here too. The enum value can be the suffix `'.waitForTaskToken'`, `'.sync'`. or `''`

src/stepfunctions/steps/utils.py
```diff
@@ -12,6 +12,46 @@
...
+            cur_partition = partition
+            return cur_partition
+
+    return cur_partition
+
+
+"""
+ARN builder for task integration
+"""
+def resource_integration_arn_builder(service, api, integration_pattern=None):
```
wong-a, line 51: 1. This belongs in the service integration module
2. Since this a method, let's name it with a verb instead. e.g. `get/build_service_integration_arn`

src/stepfunctions/steps/integration_resources.py
```diff
@@ -0,0 +1,89 @@
...
+    ECS = "ecs"
+    Batch = "batch"
+    DynamoDB = "dynamodb"
+    SNS = "sns"
+    SQS = "sqs"
+    ElasticMapReduce = "elasticmapreduce"
+
+
+class LambdaApi(Enum):
+    Invoke = "invoke"
```
wong-a, line 41: Let's remove these. We don't need to refer to the service or api names outside the Step classes.

wong-a (changes requested):
Some more comments about structure, naming, and documentation. Almost there!

src/stepfunctions/steps/utils.py
yuan-bwn: Makes sense. Will make the change.

src/stepfunctions/steps/integration_resources.py
yuan-bwn: I had similar thought before I start coding. 
But during the implementation, I found it's important to have the well organized service/api for development. It will not only help with auto-completion which helps accelerate coding speed, but will also help reading/changing/examining the code as we have our resource gathered in a place rather than sparsely distributed among places.
Another reason is: we have the integration pattern passed in as enum, this is limited number hard-coded param exactly same as our integrated services and their apis. 
I think this will also help with organizing further expansion when we are introducing more service and apis into the repo.

src/stepfunctions/steps/utils.py
```diff
@@ -12,6 +12,46 @@
...
+
 def tags_dict_to_kv_list(tags_dict):
-    kv_list = [{"Key": k, "Value": v} for k,v in tags_dict.items()]
-    return kv_list
\ No newline at end of file
+    kv_list = [{"Key": k, "Value": v} for k, v in tags_dict.items()]
+    return kv_list
+
+
+"""
```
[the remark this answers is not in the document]
yuan-bwn: Sorry I missed this point. Will update.

src/stepfunctions/steps/compute.py
```diff
@@ -37,10 +39,20 @@ def __init__(self, state_id, wait_for_callback=False, **kwargs):
             result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
+
         if wait_for_callback:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::lambda:invoke.waitForTaskToken'
+            """
+            Example resource arn: arn:aws:states:::lambda:invoke.waitForTaskToken
+            """
+            kwargs[Field.Resource.value] =  resource_integration_arn_builder(IntegrationServices.Lambda,
```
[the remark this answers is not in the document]
yuan-bwn: Will change in update

src/stepfunctions/steps/integration_resources.py
yuan-bwn: Sure. Will update

src/stepfunctions/steps/integration_resources.py
wong-a: I'm not convinced. Arguments for types and autocompletion make sense if they are to be used in many places. That is not the case for "lambda" and "invoke", etc.. There's no need to reference them outside the module defining the service integration steps, so don't expose them.

> Another reason is: we have the integration pattern passed in as enum, this is limited number hard-coded param exactly same as our integrated services and their apis.

Not sure what you mean here

src/stepfunctions/steps/utils.py
```diff
@@ -45,13 +49,19 @@ def get_aws_partition():
...
-ARN builder for task integration
-"""
-def resource_integration_arn_builder(service, api, integration_pattern=None):
+def get_service_integration_arn(service, api, integration_pattern=IntegrationPattern.RequestResponse):
+
+    """
+    ARN builder for task integration
+    Args:
+        service(str): name of the task resource service
+        api(enum): api to be integrated of the task resource service
```
wong-a, line 58:
```suggestion
        api(str): api to be integrated of the task resource service
```

src/stepfunctions/steps/utils.py
```diff
@@ -45,13 +49,19 @@ def get_aws_partition():
...
+        integration_pattern(enum, optional): integration pattern for the task resource.
```
wong-a, line 59:
```suggestion
        integration_pattern(IntegrationPattern, optional): integration pattern for the task resource.
```

src/stepfunctions/steps/utils.py
```diff
@@ -45,13 +49,19 @@ def get_aws_partition():
     return cur_partition
 
 
-"""
-ARN builder for task integration
-"""
-def resource_integration_arn_builder(service, api, integration_pattern=None):
+def get_service_integration_arn(service, api, integration_pattern=IntegrationPattern.RequestResponse):
```
wong-a, line 52: Move this to `integration_resources.py`. Having an unfocused `utils` module is usually an antipattern. You created a new module for service integration resources and this method definitely belongs there.

src/stepfunctions/steps/compute.py
```diff
@@ -12,10 +12,32 @@
...
 from stepfunctions.steps.fields import Field
-from stepfunctions.steps.utils import resource_integration_arn_builder
-from stepfunctions.steps.integration_resources import IntegrationPattern, IntegrationServices, LambdaApi, GlueApi, BatchApi, EcsApi
+from stepfunctions.steps.utils import get_service_integration_arn
+from stepfunctions.steps.integration_resources import IntegrationPattern
+
+Lambda = "lambda"
+Glue = "glue"
+Ecs = "ecs"
+Batch = "batch"
```
wong-a, line 24:
```suggestion
LAMBDA_SERVICE_NAME = "lambda"
GLUE_SERVICE_NAME = "glue"
ECS_SERVICE_NAME = "ecs"
BATCH_SERVICE_NAME = "batch"
```

nit: Variable names should written be in `snake_case`. Class names are `PascalCase`. Constants in `CAPITAL_SNAKE_CASE`. Repeats.

src/stepfunctions/steps/compute.py
```diff
@@ -12,10 +12,32 @@
...
+
+
+class LambdaApi(Enum):
```
wong-a, line 27: nit (not blocking): Not necessary to be a class

src/stepfunctions/steps/compute.py
```diff
@@ -37,10 +61,22 @@ def __init__(self, state_id, wait_for_callback=False, **kwargs):
...
+
         if wait_for_callback:
-            kwargs[Field.Resource.value] = 'arn:aws:states:::lambda:invoke.waitForTaskToken'
+            """
+            Example resource arn: arn:aws:states:::lambda:invoke.waitForTaskToken
+            """
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(Lambda,
+                                                                        LambdaApi.Invoke,
+                                                                        IntegrationPattern.WaitForTaskToken)
```
wong-a, line 72: nit: indentation is off

wong-a (changes requested):

src/stepfunctions/steps/utils.py
wong-a: (2) still applies

src/stepfunctions/steps/utils.py
yuan-bwn: Will replace "str" with "<Service>Api" as it could provide a clearer explanation

src/stepfunctions/steps/utils.py
yuan-bwn: Will do

src/stepfunctions/steps/utils.py
yuan-bwn: Makes sense. Will do.

src/stepfunctions/steps/compute.py
yuan-bwn: Will fix

src/stepfunctions/steps/utils.py
wong-a: That's more confusing as it references a type that doesn't exist. Keep as `str`. The parameter name is already `api`.

src/stepfunctions/steps/integration_resources.py
```diff
@@ -26,5 +27,22 @@ class IntegrationPattern(Enum):
...
 
+def get_service_integration_arn(service, api, integration_pattern=IntegrationPattern.RequestResponse):
+
+    """
+    ARN builder for task integration
+    Args:
+        service(str): name of the task resource service
+        api(<Service>Api): api to be integrated of the task resource service
+        integration_pattern(IntegrationPattern, optional): integration pattern for the task resource.
+                                            Default as request response.
```
wong-a, line 38: Please apply the suggestion below. You can do this in the UI by clicking the "Commit suggestion" button.
```suggestion
        service (str): The service name for the service integration
        api (str): The api of the service integration
        integration_pattern (IntegrationPattern, optional): The integration pattern for the task. (Default: IntegrationPattern.RequestResponse)
```

wong-a (approved):
LGTM. Approval conditional on applying the docstring suggestion for `get_service_integration_arn`

src/stepfunctions/steps/utils.py
yuan-bwn: So in the new commit I replaced it with "<Service>Api" which I think it is clearer than only "Api". Sorry forgot to update it here.
Let me know if that's still confusing. I can change it to "str". The reason I think "str" is not the best choice is: I implemented the API as one enum and fetch the value in the get_service_integration_arn method. "str" is not really reflecting the input.

wong-a (approved):

src/stepfunctions/steps/service.py
```diff
@@ -35,7 +67,13 @@ def __init__(self, state_id, **kwargs):
...
             output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
-        kwargs[Field.Resource.value] = 'arn:aws:states:::dynamodb:getItem'
+
+        """
+        Example resource arn: arn:aws:states:::dynamodb:getItem
+        """
+
+        kwargs[Field.Resource.value] = get_service_integration_arn(DYNAMODB_SERVICE_NAME,
+                                                                   DynamoDBApi.GetItem)
```
shivlaks, line 76: thanks for adding these through the PR, it definitely helps as a review and will benefit contributors!

src/stepfunctions/steps/compute.py
```diff
@@ -12,8 +12,31 @@
...
+GLUE_SERVICE_NAME = "glue"
+ECS_SERVICE_NAME = "ecs"
+BATCH_SERVICE_NAME = "batch"
+
+
+class LambdaApi(Enum):
+    Invoke = "invoke"
+
+
+class GlueApi(Enum):
```
shivlaks, line 30: not blocking: also agree that these don't **need** to be classes. 

If you do keep them as classes, we should include docstrings that explain what they are and their purpose :)

src/stepfunctions/steps/integration_resources.py
```diff
@@ -0,0 +1,46 @@
...
+        service (str): The service name for the service integration
+        api (str): The api of the service integration
+        integration_pattern (IntegrationPattern, optional): The integration pattern for the task. (Default: IntegrationPattern.RequestResponse)
+    """
+    arn = ""
+    if integration_pattern == IntegrationPattern.RequestResponse:
+        arn = f"arn:{get_aws_partition()}:states:::{service}:{api.value}"
+    else:
+        arn = f"arn:{get_aws_partition()}:states:::{service}:{api.value}.{integration_pattern.value}"
+    return arn
```
shivlaks, line 44: really nice to see us leveraging f-strings!

src/stepfunctions/steps/service.py
```diff
@@ -12,8 +12,40 @@
...
 from stepfunctions.steps.fields import Field
+from stepfunctions.steps.integration_resources import IntegrationPattern, get_service_integration_arn
+
+DYNAMODB_SERVICE_NAME = "dynamodb"
+SNS_SERVICE_NAME = "sns"
+SQS_SERVICE_NAME = "sqs"
+ELASTICMAPREDUCE_SERVICE_NAME = "elasticmapreduce"
+
+
+class DynamoDBApi(Enum):
```
shivlaks, line 26: all our classes should include docstrings

shivlaks (approved):
🎉 changes LGTM! provisional approval on adding doc-strings to classes without them.

Issue #98, opened 2020-10-12
Feature request: adding output_path parameter to TrainingStep constructor


lucaboulard:
Currently it is not possible to specify the output path to TrainingStep. The one specified to the wrapped estimator counts. But estimartos does not support placeholders, so the output path must be known at workflow creation time.
It would be useful to be able to specify a dynamic output path to the TrainingStep using placeholders, overriding the estimator output_path .

ca-nguyen:
Hello Luca, 
Thank you for this feature request!
Starting work on a PR to support the use of dynamic output_path for TrainingStep - it will be published in the following days.

Issue #97, opened 2020-10-12
Feature request: support placeholders in the data parameter of TrainingStep constructor


lucaboulard:
Currently the `data` parameter to the `TrainingStep` constructor does not support placeholders, requiring the s3 location of the dataset to be known at workflow creation time.
It would be useful for this parameter to accept dynamic s3 uri, via placeholders.

shivlaks:
Implemented in  #142  which has now been merged 🎉

Issue #80, opened 2020-08-25
Can't define OutputDataConfig for TrainingStep 


szamarin:
It doesn't look like it's possible to specify the output path for the model artifacts in the TrainingStep class. You can only use the default output directory or define it in the estimator object before you pass it to the TrainingStep. In either case it doesn't look like its possible to set the output path from the execution input or the output of a prior task.

a13zen:
+1 on this. 

One could potentially work around this by overwriting the sagemaker default bucket (https://github.com/aws/sagemaker-python-sdk/issues/1402), but this doesn't allow you to specify a path, only a different bucket.

As the API integration supports creating the OutputDataConfig (https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html), it makes sense to expose it via this SDK directly.

shivlaks:
Implemented in  #142  which has now been merged 🎉

Issue #139, opened 2021-06-07
Feature request: supporting placeholders in ProcessingStep constructor


lucaboulard:
Currently, the `container_arguments` parameter to the `ProcessingStep` constructor does not support placeholders, so that the only way to pass dynamic values to a processing job is to use a lambda to create a file on s3 with the values in it and bring it into the job as a processing input. Unfortunately this is quite inconvenient.
It would be a major improvement if `ProcessingStep`'s constructor accepted placeholders.

ca-nguyen:
The fix (#155) will be available with next release

yanhong-zhao-ef:
I am wondering if we can re-open this issue since the original feature request is still not possible where execution_input['argument'].to_jsonpath() will get passed straight to the processing script without the variable being replaced on execution and execution_input['argument'] runs into TypeError: Object of type ExecutionInput is not JSON serializable when trying to create workflow

PR #142, opened 2021-06-11
feat: Adds support for Placeholders in TrainingStep to set S3 location for InputDataConfig and OutputDataConfig

ca-nguyen:
*Issue #, if available:* #98, #97, #80

*Description of changes:*
Currently, it is not possible to specify the output path to TrainingStep. It needs to be defined at the Estimator level, which does not support placeholders. This change will make it possible to insert a placeholder output path from the TrainingStep definition and propagate it dynamically to the Estimator.

This change change also makes the TrainingStep data parameter compatible with placeholders.

There are other feature requests to make other TrainingStep arguments and ProcessingStep arguments compatible with placeholders. They will be addressed in a separate PR where the implementation could maybe be done at a higher level to avoid repetition.

---


src/stepfunctions/steps/sagemaker.py
```diff
@@ -103,6 +110,10 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
         if isinstance(job_name, (ExecutionInput, StepInput)):
             parameters['TrainingJobName'] = job_name
 
+        if output_path is not None:
+            if isinstance(output_path, (ExecutionInput, StepInput)):
+                parameters['OutputDataConfig']['S3OutputPath'] = output_path
```
wong-a, line 115: Looking at the [CreateTrainingJob API docs](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html) there's a lot of the API surface we aren't covering. e.g. `InputDataConfig`, `ResourceConfig`, etc. I am also see a few other parameters that accept an S3OutputPath, like [DebugRuleConfigurations](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html#sagemaker-CreateTrainingJob-request-DebugRuleConfigurations).

I think there are two approaches we can take here which applies to all Tasks:
1. Selectively expose properties can use placeholders
2. Change how Task constructs the ASL to always replace placeholders in parameters (and result_selector when we add it)

wong-a: Actually, it looks like that already happens in the JSON serialization:
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L51-L77

So the issue is more about making all arguments we expose from the TrainingStep compatible with placeholders.

ca-nguyen: Yes, agreed!
We would need to pass all TrainingStep potential placeholder arguments to parameters so that they can be replaced during JSON serialization.

shivlaks: > So the issue is more about making all arguments we expose from the TrainingStep compatible with placeholders.

yes, that's what I had in mind as well for implementation here (and related issues)

ca-nguyen: Let's focus on data and output_path for this PR and make the other arguments compatible with placeholders in another. There is a [feature request ](https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/139) to make ProcessingStep arguments compatible with placeholders as well - we could do this on a higher level (for ex [Task](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L583)) so that there is less repetitiveness.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -103,6 +116,16 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
         if isinstance(job_name, (ExecutionInput, StepInput)):
             parameters['TrainingJobName'] = job_name
 
+        if output_path is not None:
+            if isinstance(output_path, (ExecutionInput, StepInput)) or isinstance(output_path, str):
+                parameters['OutputDataConfig']['S3OutputPath'] = output_path
+
+        if data is not None and data_placeholder:
+            # Replace the 'S3Uri' key with one that supports JSONpath value.
```
wong-a, line 124: This is already done here isn't it?
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L51-L77

src/stepfunctions/steps/sagemaker.py
```diff
@@ -88,6 +95,12 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
 
             kwargs[Field.Resource.value] = get_service_integration_arn(SAGEMAKER_SERVICE_NAME,
                                                                        SageMakerApi.CreateTrainingJob)
+        # sagemaker.workflow.airflow.training_config does not accept ExecutionInput as input, only str, dict,
+        # TrainingInput or FileSystemInput. Transform placeholder into JSONpath to be processed and included in
+        # generated Sagemaker parameters. Placeholder for uri str only:
+        data_placeholder = isinstance(data, (ExecutionInput, StepInput))
```
wong-a, line 101: Prefix boolean variables with `is_` so we know it's a boolean. Additionally, we should be able to use the parent class `Placeholder` for the type check. Repeats for other places where we are checking for Placeholders.
```suggestion
        is_data_placeholder = isinstance(data, (Placeholder))
```

src/stepfunctions/steps/sagemaker.py
```diff
@@ -103,6 +116,16 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
...
             parameters['TrainingJobName'] = job_name
 
+        if output_path is not None:
+            if isinstance(output_path, (ExecutionInput, StepInput)) or isinstance(output_path, str):
+                parameters['OutputDataConfig']['S3OutputPath'] = output_path
+
+        if data is not None and data_placeholder:
+            # Replace the 'S3Uri' key with one that supports JSONpath value.
+            # Support for uri str only: The list will only contain 1 element
+            temp_data_uri = parameters['InputDataConfig'][0]['DataSource']['S3DataSource'].pop('S3Uri', None)
```
wong-a, line 126: Is [`sagemaker.workflow.airflow.training_config`](https://sagemaker.readthedocs.io/en/stable/workflows/airflow/sagemaker.workflow.airflow.html#training-config) going to have all these properties? Seems fragile to do this without any checks.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -88,6 +95,12 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
 
             kwargs[Field.Resource.value] = get_service_integration_arn(SAGEMAKER_SERVICE_NAME,
                                                                        SageMakerApi.CreateTrainingJob)
+        # sagemaker.workflow.airflow.training_config does not accept ExecutionInput as input, only str, dict,
+        # TrainingInput or FileSystemInput. Transform placeholder into JSONpath to be processed and included in
+        # generated Sagemaker parameters. Placeholder for uri str only:
```
wong-a, line 100: I didn't really understand this comment.

tests/unit/test_sagemaker_steps.py
```diff
@@ -265,6 +267,78 @@ def test_training_step_creation(pca_estimator):
...
 
 
+@patch('botocore.client.BaseClient._make_api_call', new=mock_boto_api_call)
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_training_step_creation_with_placeholders(pca_estimator):
+    execution_input = ExecutionInput(schema={
+        'Data': str,
+        'JobName': str,
+        'OutputPath': str,
+    })
```
wong-a, line 277: ExecutionInput is just one type of placeholder. Add StepInput placeholders to test that that works too.

src/stepfunctions/steps/sagemaker.py
wong-a: Nevermind. `parameters` becomes the output of [`sagemaker.workflow.airflow.training_config`](https://sagemaker.readthedocs.io/en/stable/workflows/airflow/sagemaker.workflow.airflow.html#training-config)  from L106-108 which doesn't have Placeholders.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -103,6 +116,16 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
         if isinstance(job_name, (ExecutionInput, StepInput)):
             parameters['TrainingJobName'] = job_name
 
+        if output_path is not None:
+            if isinstance(output_path, (ExecutionInput, StepInput)) or isinstance(output_path, str):
```
wong-a, line 120: This silently ignores invalid types. We should either remove this type check and let things fail downstream or throw an error if output_path is not a str or Placeholder

src/stepfunctions/steps/sagemaker.py
wong-a: So my comment now is actually about what `output_path` refers to. As I mentioned earlier, CreateTrainingJob has 6 instances of S3OutputPath in different parameters. Is `output_path` descriptive enough or should the name be more specific? I believe the current name was just chosen from https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/98

tests/integ/test_sagemaker_steps.py
```diff
@@ -104,6 +105,65 @@ def test_training_step(pca_estimator_fixture, record_set_fixture, sfn_client, sf
...
+
+        # Check placeholders in workflow definition:
+        execution = workflow.execute(inputs=execution_input)
+        expected_output_data_key_value_pair = ("S3OutputPath.$", "$$.Execution.Input['OutputPath']")
+        assert expected_output_data_key_value_pair in \
+               workflow.definition.states[training_step_name]['Parameters']['OutputDataConfig'].items()
+
+        expected_job_name_key_value_pair = ("TrainingJobName.$", "$$.Execution.Input['JobName']")
+        assert expected_job_name_key_value_pair in \
+               workflow.definition.states[training_step_name]['Parameters'].items()
```
wong-a, line 157: Static ASL validation is already covered in the unit tests and doesn't require an execution. `create_workflow_and_check_definition` already asserts that the created definition from DescribeStateMachine matches the python object.

tests/integ/test_sagemaker_steps.py
```diff
@@ -104,6 +105,65 @@ def test_training_step(pca_estimator_fixture, record_set_fixture, sfn_client, sf
...
+        # Check workflow output
+        execution_output = execution.get_output(wait=True)
+        assert execution_output.get("TrainingJobStatus") == "Completed"
+
+        # Cleanup
+        state_machine_delete_wait(sfn_client, workflow.state_machine_arn)
+        # End of Cleanup
```
wong-a, line 164: What's the value in this integ test? It doesn't make any assertions that the placeholders provided are actually resolved during execution

wong-a (changes requested):

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Yes, it is. For this case, to generate the training parameters, we need to call [sagemaker.workflow.airflow.training_config](https://sagemaker.readthedocs.io/en/stable/workflows/airflow/sagemaker.workflow.airflow.html) with _inputs=data_ which does not accept [ExecutionInput](https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=executionInput#stepfunctions.inputs.ExecutionInput) or [StepInput](https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=executionInput#stepfunctions.inputs.StepInput) objects for _inputs_. To bypass that, if _data_ is either an [ExecutionInput](https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=executionInput#stepfunctions.inputs.ExecutionInput) or [StepInput](https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=executionInput#stepfunctions.inputs.StepInput) object, we transform it to _JSONpath_ before calling [training_config()](https://sagemaker.readthedocs.io/en/stable/workflows/airflow/sagemaker.workflow.airflow.html) (which will be processed as a string). Doing that, the parameters will be generated with the JSONpath as uri, but the corresponding key won't be formatted to take JSONpath value (it will be missing the ".$" at the end).

Alternatively, we could also modify [_replace_placeholders()](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L51-L65) to replace the key (and add ".$") when the value is already a JSONpath. That could be a cleaner way to do it and could be re-used as we make more parameters placeholder compatible.

wong-a: Yes, I figured that out myself after reading more of the code.

> Alternatively, we could also modify _replace_placeholders() to replace the key (and add ".$") when the value is already a JSONpath. That could be a cleaner way to do it and could be re-used as we make more parameters placeholder compatible.

If you're talking about replacing the keys regular strings which happen to be JSONPath strings, we can't do that because it's backwards incompatible. It's valid usage to pass a normal string `"$.that.is.jsonpath"` as a value for any parameter.

tests/unit/test_sagemaker_steps.py
ca-nguyen: Good point! Will add a test case for StepInput in the next commit!

src/stepfunctions/steps/sagemaker.py
ca-nguyen: +1
Agreed! Let's keep it that way and maybe add checks to make sure the accessed keys in parameters exist.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Yes, with the current Sagemaker version we are using, these properties are added whenever with call the function with a str type for _inputs_. Agreed, will add checks to make it more robust and in case Sagemaker makes changes for the format.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agreed!
Will make the changes in the next commits.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agreed. This comment definitely need to be clearer and more concise! Will do in the next commit(s)!

tests/integ/test_sagemaker_steps.py
ca-nguyen: You are right - will remove the asserts!

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Right, we definitely don't want to ignore those! Will remove the checks.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: The defined parameters are passed down to the [sagemaker.estimator.EstimatorBase](https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html?highlight=EstimatorBase#sagemaker.estimator.EstimatorBase). `output_path` name was used here to concord with the `output_path` in [EstimatorBase](https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html?highlight=EstimatorBase#sagemaker.estimator.EstimatorBase). I feel it might add confusion to name it something else, but I am not strongly opinionated about it - we could use a more descriptive name and mention in the docs that it maps to [EstimatorBase](https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html?highlight=EstimatorBase#sagemaker.estimator.EstimatorBase)'s `output_path`.

Would `result_path` be a better option?

ca-nguyen: Looking at the [CreateTrainingJob](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateTrainingJob.html) docs you mentioned earlier, would `output_data_config_path` work? Since `OutputDataConfig` is the property we update in the parameters, this could be a better option.

tests/integ/test_sagemaker_steps.py
ca-nguyen: This was added to make sure the workflow executed without errors, but I agree that it does not validate that the placeholders were replaced with the execution input values. Would the unit test suffice? If so, I will remove this test.

src/stepfunctions/steps/sagemaker.py
wong-a: nit: Can just call it `data_uri`. I don't think there's anything "temporary" about it - there's no existing `data_uri` variable and it's not modified either

tests/unit/test_sagemaker_steps.py
wong-a: You can combine them in this test case:

```python
        'Data': str,
        'OutputPath': str,
    })

    step_input = ExecutionInput(schema={
        'JobName': str,
    })

   # ...
   job_name=step_input['JobName']

   #... assert
   'TrainingJobName.$': ...
```

src/stepfunctions/steps/sagemaker.py
wong-a: `output_data_config_path` seems alright to me

Both `result_path` and `output_path` conflict with existing Task properties of the same name, so we can't use those. Sagemaker steps extend `Task` and the caller should still be able to specify those fields like any other Task.
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/6091932448572244801362419034414b48c9f4e6/src/stepfunctions/steps/states.py#L601-L602

tests/integ/test_sagemaker_steps.py
wong-a: Yes, I think a unit test is good enough. You *can* check what happened at runtime by making assertions on the execution history though.

wong-a (changes requested):
Almost there! Remaining items:

- [x] Update name of `output_path` argument to not clash with existing `output_path` for `Task`
- [x] Comment and variable naming improvements
- [ ] Unit tests

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agreed! `data_uri` better describes the variable

tests/unit/test_sagemaker_steps.py
ca-nguyen: Good idea! Will do in the next commit

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agreed - `result_path` and `output_path` should not be used if they are conflicting with Task properties. 
`output_path` will be renamed to `output_data_config_path` in the next commit.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -88,6 +95,11 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
 
             kwargs[Field.Resource.value] = get_service_integration_arn(SAGEMAKER_SERVICE_NAME,
                                                                        SageMakerApi.CreateTrainingJob)
+        # sagemaker.workflow.airflow.training_config does not accept Placeholder as input. Transform data placeholder
+        # to JSONpath to generate parameters.
```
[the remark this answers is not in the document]
ca-nguyen: Agreed! Will add the changes in the next commit

tests/integ/test_sagemaker_steps.py
ca-nguyen: Removed integ test

tests/integ/test_sagemaker_steps.py
```diff
@@ -29,7 +29,7 @@
 from sagemaker.tuner import HyperparameterTuner
 from sagemaker.processing import ProcessingInput, ProcessingOutput
 
-from stepfunctions.inputs import ExecutionInput
+from stepfunctions.inputs import ExecutionInput, StepInput
```
wong-a, line 32:
```suggestion
from stepfunctions.inputs import ExecutionInput
```

src/stepfunctions/steps/sagemaker.py
```diff
@@ -115,14 +118,20 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
...
 
         if data is not None and is_data_placeholder:
             # Replace the 'S3Uri' key with one that supports JSONpath value.
             # Support for uri str only: The list will only contain 1 element
-            temp_data_uri = parameters['InputDataConfig'][0]['DataSource']['S3DataSource'].pop('S3Uri', None)
-            parameters['InputDataConfig'][0]['DataSource']['S3DataSource']['S3Uri.$'] = temp_data_uri
+            try:
+                data_uri = parameters['InputDataConfig'][0]['DataSource']['S3DataSource'].pop('S3Uri', None)
+                parameters['InputDataConfig'][0]['DataSource']['S3DataSource']['S3Uri.$'] = data_uri
+            except KeyError as error:
```
wong-a, line 130: This could also result in an `IndexError` if `InputDataConfig` is empty. My concern was if `training_config` always returns a structure like this. If it's guaranteed getting here actually is exceptional and we shouldn't swallow the error unless we can handle it somehow.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -69,6 +71,14 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
...
             wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait for the training job to complete before proceeding to the next step in the workflow. Set to `False` if the Task state should submit the training job and proceed to the next step. (default: True)
             tags (list[dict], optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
+            output_data_config_path (str or Placeholder, optional): S3 location for saving the training result (model
+                artifacts and output files) to propagate to estimator (maps to sagemaker.estimator.EstimatorBase
+                output_path parameter.) If not specified, results are
+                stored to a default bucket. If the bucket with the specific name
+                does not exist, the estimator creates the bucket during the
+                :meth:`~sagemaker.estimator.EstimatorBase.fit` method execution.
+                file:// urls are used for local mode. For example: 'file://model/'
+                will save to the model folder in the current directory.
```
wong-a, line 81: Reading this again, I don't think the description is true. If not specified, `OutputDataConfig` uses the `output_path` from the `estimator`. This new parameter is to override `estimator`'s output_path.

```suggestion
            output_data_config_path (str or Placeholder, optional): S3 location for saving the training result (model
                artifacts and output files). If specified, it overrides the `output_path` property of `estimator`.
```

wong-a (changes requested):

src/stepfunctions/steps/sagemaker.py
ca-nguyen: It is guaranteed as long as the structure returned by [sagemaker.workflow.airflow.training_config](https://sagemaker.readthedocs.io/en/stable/workflows/airflow/sagemaker.workflow.airflow.html) stays the same.

ca-nguyen: I'll remove the try catch so that the error is not swallowed when the structure is changed

tests/integ/test_sagemaker_steps.py
ca-nguyen: Will remove both ExecutionInput and StepInput in next commit (removed the integ test with placeholders so ExecutionInput isn't used anymore)

src/stepfunctions/steps/sagemaker.py
```diff
@@ -24,7 +26,7 @@
 from sagemaker.model_monitor import DataCaptureConfig
 
 SAGEMAKER_SERVICE_NAME = "sagemaker"
-
+logger = logging.getLogger('stepfunctions.sagemaker')
```
wong-a, line 29: Try to pay attention to when you have unused variables. Your IDE should tell you this. Since you removed the logging in your last commit, this is unused now.

However, https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/144 also added logging so you just need to resolve the merge conflict after merging from main.

https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L28

ca-nguyen: Agreed!
The logger was removed in the latest commit

src/stepfunctions/steps/sagemaker.py
```diff
@@ -15,7 +15,9 @@
 import logging
 
 from enum import Enum
-from stepfunctions.inputs import ExecutionInput, StepInput
+import logging
```
ca-nguyen, line 18: To be removed in next commit

wong-a (approved):

shivlaks (approved):

Issue #72, opened 2020-08-04
Tensorflow estimator script mode not working if we pass HyperParameters as part of TrainingStep


Axe129:
Hello,

I was trying to use Tensorflow estimator in script mode and I was trying to pass the hyper-parameters as part of `stepfunctions.steps.TrainingStep` as follows: 
<img width="752" alt="Screenshot 2020-08-04 at 6 22 01 PM" src="https://user-images.githubusercontent.com/12937657/89296024-7330f380-d67f-11ea-8475-69c305a90be8.png">

But, the training job failed with the following error
<img width="960" alt="Screenshot 2020-08-04 at 6 11 50 PM" src="https://user-images.githubusercontent.com/12937657/89295023-fa7d6780-d67d-11ea-94ec-0867d003a39b.png">

Upon investigation I found out that other hyper-parameters that the estimator attaches on its own were not going (ex - 'sagemaker_program', 'sagemaker_submit_directory' etc.)

For example, the following image is the set of hyper-parameters for the job when we give it as part of TrainingStep:
<img width="645" alt="Screenshot 2020-08-04 at 6 21 42 PM" src="https://user-images.githubusercontent.com/12937657/89296055-7b892e80-d67f-11ea-8b3b-f8a419aca847.png">

But, if we give the hyper-parameters as part of the estimator: 
<img width="1029" alt="Screenshot 2020-08-04 at 6 12 59 PM" src="https://user-images.githubusercontent.com/12937657/89295741-07e72180-d67f-11ea-824e-e381444bac00.png">

The reason why I'd want to give hyper-parameters as part of the TrainingStep is that I'd be able to use `stepfunctions.inputs.ExecutionInput` in order to provide inputs dynamically that might be useful in the script.

shivlaks:
fixed through #144 which enables the capability to specify hyperparameters in the TrainingStep constructor and merges them with hyperparameters established in the estimator.

PR #144, opened 2021-06-14
fix: supplying hyperparameters to training step constructor drops hyperparameters specified in estimator

### LINKED ISSUE #99: Enhancement: TrainingStep hyperparameters updating the estimator ones
Currently if a dict is passed to the hyperparameters param in TrainingStep constructor, this dict replaces the estimator one.
This way, not only all hyperparameters passed to the estimator constructor are removed, but the reserved ones (`sagemaker_*` ) are too.
So currently, in order to use hyperparameters in TrainingStep, you have to figure out which reserved hyperparams are required for the job to work and provide them explicitely.
It would be useful for the hyperparams passed to TrainingStep to update, rather than replace, the estimator hyperpars' dict.

shivlaks:
### Summary

`Hyperparameters` can be specified in the `estimator` object and `hyperparameters` property.
Both of which are taken in the constructor of the `TrainingStep` class.

The current behaviour drops any hyperparameters that were specified in the estimator if the property
is set in the `TrainingStep` constructor. This is undesirable as the estimators often specify algorithm specific hyperparameters out of the box that we don't want to drop.

This change merges the hyperparameters in the constructor as well as the estimator that is used in `TrainingStep`.
If there are duplicate keys, the hyperparameters specified in the constructor will be used.

Closes #99, #72

---

### DIFF
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -68,7 +68,9 @@ class TrainingStep(Task):
                 * (list[sagemaker.amazon.amazon_estimator.RecordSet]) - A list of
                     :class:`sagemaker.amazon.amazon_estimator.RecordSet` objects,
                     where each instance is a different channel of training data.
-            hyperparameters (dict, optional): Specify the hyperparameters that are set before the model begins training. If hyperparameters provided are also specified in the estimator, the provided value will used. (Default: Hyperparameters specified in the estimator will be used for training.)
+            hyperparameters (dict, optional): Parameters used for training.
+                    Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
+                    If there are duplicate entries, the value provided through this property will be used. (Default: Hyperparameters specified in the estimator.)
             mini_batch_size (int): Specify this argument only when estimator is a built-in estimator of an Amazon algorithm. For other estimators, batch size should be specified in the estimator.
             experiment_config (dict, optional): Specify the experiment config for the training. (Default: None)
             wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait for the training job to complete before proceeding to the next step in the workflow. Set to `False` if the Task state should submit the training job and proceed to the next step. (default: True)
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -104,7 +104,11 @@ class TrainingStep(Task):
             parameters['TrainingJobName'] = job_name
 
         if hyperparameters is not None:
-            parameters['HyperParameters'] = hyperparameters
+            merged_hyperparameters = {}
+            if estimator.hyperparameters() is not None:
+                merged_hyperparameters.update(estimator.hyperparameters())
+            merged_hyperparameters.update(hyperparameters)
+            parameters['HyperParameters'] = merged_hyperparameters
 
         if experiment_config is not None:
             parameters['ExperimentConfig'] = experiment_config
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -12,6 +12,8 @@
 # permissions and limitations under the License.
 from __future__ import absolute_import
 
+import logging
+
 from enum import Enum
 from stepfunctions.inputs import ExecutionInput, StepInput
 from stepfunctions.steps.states import Task
@@ -22,6 +24,8 @@ from stepfunctions.steps.integration_resources import IntegrationPattern, get_se
 from sagemaker.workflow.airflow import training_config, transform_config, model_config, tuning_config, processing_config
 from sagemaker.model import Model, FrameworkModel
 from sagemaker.model_monitor import DataCaptureConfig
+
+logger = logging.getLogger('stepfunctions.sagemaker')
 
 SAGEMAKER_SERVICE_NAME = "sagemaker"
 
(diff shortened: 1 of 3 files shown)


src/stepfunctions/steps/sagemaker.py
```diff
@@ -104,7 +104,11 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
             parameters['TrainingJobName'] = job_name
 
         if hyperparameters is not None:
-            parameters['HyperParameters'] = hyperparameters
+            merged_hyperparameters = {}
+            if estimator.hyperparameters() is not None:
+                merged_hyperparameters.update(estimator.hyperparameters())
```
wong-a, line 109: re:
>  should we warn/error instead of an in-place merge when we find duplicate keys?

Throwing an error wouldn't be helpful; it's more breaking (you can't create what you could before) and doesn't address the desired functionality in #99. Logging might at an INFO level may be helpful for duplicate keys.

We should also document if not already that `parameters` gets totally overridden by `training_config`. This isn't the case for all service integration steps. Maybe we should adopt an update strategy there too?

tests/unit/test_sagemaker_steps.py
```diff
@@ -482,6 +482,135 @@ def test_training_step_creation_with_framework(tensorflow_estimator):
         'End': True
     }
 
+@patch('botocore.client.BaseClient._make_api_call', new=mock_boto_api_call)
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def training_step_merges_hyperparameters_from_constructor_and_estimator(tensorflow_estimator):
```
wong-a, line 487: Just unit tests seem sufficient to me. This mostly has to do with static ASL generation and doesn't really rely on execution.

shivlaks: yeah i also deferred writing tests because it already takes long as it is and we are validating against the generated definition

src/stepfunctions/steps/sagemaker.py
shivlaks: I was originally thinking that we should log it as a warning since `INFO` tends to generally also include a lot of junk but not strongly opinionated. 

I'll re-spin the `update` calls to assemble the dicts so that they log something on duplicate keys.
absolutely agree that we need to document.

> This isn't the case for all service integration steps. Maybe we should adopt an update strategy there too?

great call. wasn't on my radar, but I'm in favour of adopting the update strategy. The most consistent it is across things in the SDK, the more intuitive and idiomatic it will feel for users.

shivlaks: I'll make the changes to service integration steps in a separate PR. let me know if you had a different thought/idea of where we should be documenting this behaviour @wong-a

wong-a: I meant the other way around actually. In all service integration steps besides sagemaker, the constructor accepts a `parameters` argument that becomes `Parameters` .  We can update sagemaker step classes to accept a `parameters` dict which can be an escape hatch for full API coverage or override any explicitly exposed arguments.

For example, DynamoDBUpdateItem the caller must specify all parameters in the `parameters` field. The constructor doesn't have a `table_name` argument to set or other required fields:
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/service.py#L140-L167

Whereas the sagemaker steps always construct parameters using the sagemaker SDK and some special arguments in the constructor. You could provide `parameters` because the constructor accepts `**kwargs`, but it won't do anything.
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L477-L479

src/stepfunctions/steps/sagemaker.py
```diff
@@ -64,7 +68,7 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
                 * (list[sagemaker.amazon.amazon_estimator.RecordSet]) - A list of
                     :class:`sagemaker.amazon.amazon_estimator.RecordSet` objects,
                     where each instance is a different channel of training data.
-            hyperparameters (dict, optional): Specify the hyper parameters for the training. (Default: None)
+            hyperparameters (dict, optional): Specify the hyperparameters that are set before the model begins training. If hyperparameters provided are also specified in the estimator, the provided value will used. (Default: Hyperparameters specified in the estimator will be used for training.)
```
wong-a, line 71: > If hyperparameters provided are also specified in the estimator, the provided value will used.

The code comment you wrote about merging was a clearer description. I don't think this sentence communicates that the `hyperparameter` arg values are merged into the estimator's hyperparameters.


---

As for other documentation, I think the docstring suffices since that shows up in RTD.

wong-a (changes requested):
LGTM besides making documentation clearer

src/stepfunctions/steps/sagemaker.py
shivlaks: gotcha, i'd taken 2 stabs and was waffling back and forth on which one I understood better. will go with the other blurb

src/stepfunctions/steps/sagemaker.py
shivlaks: I see what you mean. I'm in favour of adding that parameters property and will address it in another PR. Escape hatches are powerful because it'll give users a path forward without requiring first class support to be developed and released.

wong-a: Yeah, that's out of scope of this PR. Can you create an issue for tracking?

wong-a (dismissed):
```
< ship it >
 ---------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ca-nguyen (approved):

ca-nguyen (approved):

Issue #117, opened 2021-02-08
add placeholders support for all properties 


joschaharpeng1337:
First: Thanks for this api. Its really a handy way to create workflows.

## Problem
We tried use an execution input for the `image_uri` in a Processor. The workflow ended up using the execution_input variable as a String for image_uri. It did not replace it.

In general we want to set a lot of properties via a configuration, without changing the workflow.

## Here is what we did
```
...
execution_input = ExecutionInput()
groot_processor = Processor(
 ...
    image_uri=execution_input['image_uri']"
 ...

)
groot_step = ProcessingStep('groot', ...)
chain = Chain([groot_step])
workflow = Workflow(..., definition=chain,..)
state_machine_arn = workflow.create()

config = {
   "image_uri": "1234567890.dkr.ecr.eu-central-1.amazonaws.com/solrizer:latest"
}

import boto3
SFN_CLIENT = boto3.client('stepfunctions', region_name=boto3.Session().region_name)
SFN_CLIENT.start_execution(
        stateMachineArn=state_machine_arn,
        name='test-pipeline',
        input=json.dumps(config)
    )
...
```
Sadly the workflow did not replace `execution_input['image_uri']` with the actual value.

## This is our workaround
```
sf_client = boto_session.client('stepfunctions')
execution_input = ExecutionInput()
groot_processor = Processor(
 ...
    image_uri="$$.Execution.Input['image_uri']",
...
)
groot_step = ProcessingStep('groot', ...)
chain = Chain([groot_step])
workflow = Workflow(..., definition=chain,..)
workflow_arn = workflow.create()

response = sf_client.describe_state_machine(
    stateMachineArn=workflow_arn
)

original_workflow = json.loads(response['definition'])
for step_name in original_workflow["States"]:
    try:
        original_workflow["States"][step_name]["Parameters"]["AppSpecification"]["ImageUri.$"] = \
        original_workflow["States"][step_name]["Parameters"]["AppSpecification"].pop("ImageUri")
    except KeyError:
        continue

new_workflow = json.dumps(original_workflow, indent=2)

sf_client.update_state_machine(
    stateMachineArn=workflow_arn,
    definition=new_workflow,
    roleArn=workflow_role
)
```

used dependencies:
```
sagemaker==2.19.0
stepfunctions==2.0.0rc1
```

## Expectation
[...]

Issue #94, opened 2020-10-06
Feature Request: Support for Placeholders for Sagemaker Processing container arguments


manshahe:
This is a feature request for the newly added **stepfunctions.steps.sagemaker.ProcessingStep** to accept placeholders for inputs to the container_argument parameter. An example case is depicted below:

```
execution_input = ExecutionInput(
    schema={
        "JobName": str,
        #This argument is for the Sagemaker Processing step
        "argument": str
    }
)

processing_step = ProcessingStep(
    job_name=execution_input["JobName"],
    processor={sagemaker processor object here},
    inputs=inputs,
    outputs = outputs,
    #Argument passed here to Sagemaker Processing
    container_arguments=['--argument', execution_input.get['argument'].to_jsonpath()],
    container_entrypoint=["python3", "/opt/ml/processing/input/code/scipt.py"]
)
```

**Rationale:** This will make the generated state machines more generic, we can generate 1 state machine and pass it arguments for different workloads. Without this functionality, I need to generate **_n_** state machines for a parameter which takes **_n_** possible values.

wong-a:
Creating an array with placeholder values for specific elements will require using the `States.Array` [intrinsic function](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html) unless the referenced JSONPath value is an array.

```json
"AppSpecification.ContainerArguments.$": "States.Array('--argument', $.argument)"
```

ca-nguyen:
The fix to support placeholders (#155) will be available in next release.

yanhong-zhao-ef:
I am wondering if we can re-open this issue since the original feature request is still not possible where execution_input['argument'].to_jsonpath() will get passed straight to the processing script without the variable being replaced on execution and execution_input['argument'] runs into TypeError: Object of type ExecutionInput is not JSON serializable when trying to create workflow

PR #154, opened 2021-08-10
[DRAFT] feature: Support placeholders for Sagemaker parameters

ca-nguyen:
*Issue #, if available:* #117, #139, #94

*Description of changes:*
Currently, it is not possible to use placeholders for Sagemaker properties.
This change makes it possible to use placeholders and define SageMaker properties dynamically, after the state machine has been created (path to those params are defined in src/stepfunctions/steps/constants.py.)

With this change, it will be possible to define Sagemaker properties by passing them directly to the Sagemaker step constructor as additional args. The args that can be defined will depend on the type of step that you create.

*TODO*
- [ ] Add support VpcConfig for TrainingStep to pass to Estimator
- [ ] Add support for CheckpointConfig for TrainingStep to pass to Estimator
- [ ] Add support for InputDataConfig for TrainingStep to pass to Estimator
- [ ] Add tests for TuningStep placeholders
- [ ] Add test to catch InvalidPathToPlaceholderParameter exception
- [ ] Update doc

*Open questions*
- We parse through kwargs for all SageMaker step tasks and all valid SageMaker properties are treated to make placeholder compatible. This could be confusing as depending on the type of step, the properties can be propagated to a processor, estimator or tuner. Is passing the properties in the kwargs simple to understand/use or is there a need to separate SageMaker properties per step type?
- Should we add documentation for each configurable SageMaker parameters that can be passed down to the processor, estimator and tuner in the functions Args section? Would a sentence mentioning those configurable parameters suffice?


shivlaks (commented):
still reviewing, but i'll mention a high level thought

is it possible to lay out the approach itself as a part of the description.

What I'd be looking for is a summary of:
1. approach we have taken to support this
2. alternatives
3. open questions

I'd also say it's far more pragmatic to scaffold these out for each of the steps in separate PRs rather than all in one. It will be simpler to review as well as author. The collection of PRs can collectively resolve the larger issue of supporting "all" placeholders.

wdyt?

ca-nguyen:
I agree, this will be clearer and easier to review if each step had each own PR.
I'll break it into smaller ones

Issue #85, opened 2020-09-15
Estimator Output Path cannot accept Execution Input placeholder 


lgimyuen:
The arguments of  `sagemaker.estimator.Estimator()`  and `sagemaker.tuner.HyperparameterTuner()` doesn't accept `ExecutionInput` or `StepInput` as placeholder. E.g. for the `output_path` of `sagemaker.estimator.Estimator()` has to be fixed in the definition of workflow. 

The current way around this is not to make use of `stepfunctions.steps.sagemaker.TuningStep` and use the generic `step functions.steps.Task` to invoke the `arn:aws:states:::sagemaker:createHyperParameterTuningJob.sync` with parameters from ExecutionInput and StepInput

jackmuskopf:
Same issue with ProcessingJobs

jackmuskopf:
For instance, how do we do something like this using the `stepfunctions.steps.ProcessingStep` class?

>         "ProcessingOutputConfig": {
          "Outputs": [
            {
              "OutputName": "CleanData",
              "S3Output": {
                "S3Uri.$": "States.Format('s3://my-bucket/my/data/path/clean/{}', $$.Execution.Input['ExecutionName'])",
                "LocalPath": "/opt/ml/processing/output",
                "S3UploadMode": "EndOfJob"
              }
            }
          ]
        }

ca-nguyen:
Hi @jackmuskopf, thank you for your feature request!
Using placeholders with Sagemaker steps has been a popular requested feature - will start work on making TuningStep Placeholder compatible shortly.

> Same issue with stepfunctions.steps.ProcessingStep

Support to use placeholders with ProcessingStep has been merged and will be added with the next release (see [commit](https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/01e18c3777554882d1efc1410ecdcf25c4ec44e8))

The DataScience SDK does not currently support [StepFunctions Intrinsic Functions](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-intrinsic-functions.html), but with this change, it will possible to configure any property in the `parameters` arg with Placeholders.

For example:
```
parameters = {
   "ProcessingOutputConfig": {
      "Outputs": [
        {
          "OutputName": "CleanData",
          "S3Output": {
            "S3Uri": execution_input['s3_uri'],
            "LocalPath": "/opt/ml/processing/output",
            "S3UploadMode": "EndOfJob"
          }
        }
      ]
[...]

PR #155, opened 2021-08-12
Support placeholders for processing step

ca-nguyen:
*Issue #, if available:* #117, #139, #94

***Description of changes:***
Currently, it is not possible to use placeholders for Sagemaker Processor properties . The properties cannot be defined dynamically, as they need to be defined in the Processor which does not accept placeholders.
This change makes it possible to use placeholders for Processor properties by using the `parameters` field that are passed down from the ProcessingStep.

***Proposed changes***
Use the `parameters` field that is compatible with placeholders to define ProcessingStep properties.
Merge the [sagemaker generated configs](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L521) with the input `parameters`:
- The input `parameters` will overwrite the sagemaker generated configs if the properties were defined in both
- All ProcessingStep properties will be placeholder compatible 


src/stepfunctions/steps/constants.py
```diff
@@ -0,0 +1,30 @@
...
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+from enum import Enum
+from stepfunctions.steps.fields import Field
+
+# Path to SageMaker placeholder parameters
+placeholder_paths = {
+    # Paths taken from https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html
```
shivlaks, line 18: question: 

1. what does the comment mean when it says "taken" from the documentation link?
2. can all [properties](https://docs.aws.amazon.com/step-functions/latest/dg/connect-sagemaker.html) be represented by placeholders or is it only some?

src/stepfunctions/steps/constants.py
```diff
@@ -0,0 +1,30 @@
...
+    'ProcessingStep': {
+        Field.Role.value: ['RoleArn'],
+        Field.ImageUri.value: ['AppSpecification', 'ImageUri'],
+        Field.InstanceCount.value: ['ProcessingResources', 'ClusterConfig', 'InstanceCount'],
+        Field.InstanceType.value: ['ProcessingResources', 'ClusterConfig', 'InstanceType'],
+        Field.Entrypoint.value: ['AppSpecification', 'ContainerEntrypoint'],
+        Field.VolumeSizeInGB.value: ['ProcessingResources', 'ClusterConfig', 'VolumeSizeInGB'],
+        Field.VolumeKMSKey.value: ['ProcessingResources', 'ClusterConfig', 'VolumeKmsKeyId'],
+        Field.Env.value: ['Environment'],
+        Field.Tags.value: ['Tags'],
```
shivlaks, line 28: thought: is there a way to read these from the SDK or automate it? having this hand-rolled can be problematic for a few reasons:

1. drift if the API signatures expand
2. prone to error, since it's reliant on everything being hand rolled

src/stepfunctions/steps/constants.py
```diff
@@ -0,0 +1,30 @@
...
+placeholder_paths = {
+    # Paths taken from https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html
+    'ProcessingStep': {
+        Field.Role.value: ['RoleArn'],
+        Field.ImageUri.value: ['AppSpecification', 'ImageUri'],
+        Field.InstanceCount.value: ['ProcessingResources', 'ClusterConfig', 'InstanceCount'],
+        Field.InstanceType.value: ['ProcessingResources', 'ClusterConfig', 'InstanceType'],
+        Field.Entrypoint.value: ['AppSpecification', 'ContainerEntrypoint'],
+        Field.VolumeSizeInGB.value: ['ProcessingResources', 'ClusterConfig', 'VolumeSizeInGB'],
+        Field.VolumeKMSKey.value: ['ProcessingResources', 'ClusterConfig', 'VolumeKmsKeyId'],
```
shivlaks, line 26: curious: haven't played much with this, but do all nested properties (i.e. VolumeSizeInGb) support placeholders to supply their value? - Admittedly, Ive only supplied top level properties and haven't tinkered deep enough.

src/stepfunctions/steps/fields.py
```diff
@@ -59,10 +59,23 @@ class Field(Enum):
...
+    ImageUri            = 'image_uri'
+    InstanceCount       = 'instance_count'
+    InstanceType        = 'instance_type'
+    Entrypoint          = 'entrypoint'
+    VolumeSizeInGB      = 'volume_size_in_gb'
+    VolumeKMSKey        = 'volume_kms_key'
+    OutputKMSKey        = 'output_kms_key'
+    MaxRuntimeInSeconds = 'max_runtime_in_seconds'
+    Env                 = 'env'
+    Tags                = 'tags'
```
shivlaks, line 81: thought: is this the right place for storing these properties? everything else in this file is specific to states  and ASL, but this introduces properties specific to a service integration's API. 

some properties will also be duplicated across APIs/Service Integrations (things like role, tags, etc are probably used in multiple APIs)

another thing to think about:

as a customer, would it be more intuitive to have something like `Placeholders.SagemakerProcessingStep.blah` than `Fields.blah`

src/stepfunctions/steps/sagemaker.py
```diff
@@ -473,13 +576,19 @@ def __init__(self, state_id, tuner, job_name, data, wait_for_completion=True, ta
         super(TuningStep, self).__init__(state_id, **kwargs)
 
 
-class ProcessingStep(Task):
+class ProcessingStep(SageMakerTask):
```
shivlaks, line 579: breaking change??

the class [constructor]((https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/sagemaker.html?highlight=ProcessingStep#stepfunctions.steps.sagemaker.ProcessingStep)) used to take in a Task but now that's been changed. when customers upgrade versions, won't their existing code fail?

we cannot make breaking changes as we need to follow semantic versioning while releasing minor version updates.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -473,13 +576,19 @@ def __init__(self, state_id, tuner, job_name, data, wait_for_completion=True, ta
...
 
     """
     Creates a Task State to execute a SageMaker Processing Job.
+
+    The following properties can be passed down as kwargs to the sagemaker.processing.Processor to be used dynamically
+    in the processing job (compatible with Placeholders): role, image_uri, instance_count, instance_type,
+    volume_size_in_gb, volume_kms_key, output_kms_key
```
shivlaks, line 586: curious: what's the source of truth here? how did we verify that these properties are the ones supported.

tests/integ/test_sagemaker_steps.py
```diff
@@ -352,3 +354,85 @@ def test_processing_step(sklearn_processor_fixture, sagemaker_session, sfn_clien
...
+        # Execute workflow
+        execution = workflow.execute(inputs=execution_input)
+        execution_output = execution.get_output(wait=True)
+
+        # Check workflow output
+        assert execution_output.get("ProcessingJobStatus") == "Completed"
+
+        # Cleanup
+        state_machine_delete_wait(sfn_client, workflow.state_machine_arn)
+        # End of Cleanup
```
shivlaks, line 438: nit: i think the code is self explanatory. we can drop this comment 😅

tests/unit/test_sagemaker_steps.py
```diff
@@ -962,3 +964,117 @@ def test_processing_step_creation(sklearn_processor):
...
+                        }
+                    }
+                ]
+            },
+            'ProcessingResources': {
+                'ClusterConfig': {
+                    'InstanceCount.$': "$$.Execution.Input['instance_count']",
+                    'InstanceType.$': "$['instance_type']",
+                    'VolumeKmsKeyId.$': "$$.Execution.Input['volume_kms_key']",
+                    'VolumeSizeInGB.$': "$$.Execution.Input['volume_size_in_gb']"
```
shivlaks, line 1070: love the variety in this test to use different types of placeholders.

shivlaks (changes requested):
i think this helps illustrate the idea you have in mind - thank you again for shrinking down the PR size.

Wanted to post the comments so far so you can consider them. I still have to take a closer look at the `sagemaker.py` class as well as your open questions.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: This changes ProcessingStep's base class, but not the constructor arguments. With this change, instead of calling Task's constructor directly in __init__(), we call SageMakerTask's constructor which in turn calls Task's constructor. 

Before: 
- ProcessingtStep.__init__()
   - Task.__init__()

After:
- ProcessingtStep.__init__()
   - SageMakerTask.__init__()
      - Task.__init__()

tests/integ/test_sagemaker_steps.py
ca-nguyen: You're right! i'll remove the comments :)
They are included in all the other tests - will do a cleanup for the other tests in another PR

tests/unit/test_sagemaker_steps.py
ca-nguyen: Thanks! Included all args that we are making placeholder compatible :)

src/stepfunctions/steps/sagemaker.py
ca-nguyen: The ones that are made to be placeholder compatible are :

1. The args that are documented as being placeholder compatible in the Args section (for ex: job_name)
2. The ones that are included in placeholder_paths (src/stepfunctions/steps/constants.py)

Since we are replacing the Placeholders with the ExecutionInput when starting the job, when the Sagemaker job starts, all placeholders are replaced. If some args that we configured to hold placeholder in our state machine were not replaced, this should trigger an error. 

I can add a test to confirm the behaviour in the next commit

Thanks for bringing this up - this confirms that this documentation is not clear and we might want to switch to the **Alternative solution** where we would add all Placeholder compatible properties as optional args in the step constructor, making it clearer to the customer which are Placeholder compatible.

src/stepfunctions/steps/fields.py
ca-nguyen: Love what you are proposing! Separating the SageMaker property fields and the state and ASL specific fields will definitely make it more intuitive to the customer

Changes will be made in next commit

wong-a:
> Currently, only dicts, str and number objects were made placeholder compatible to facilitate testing purposes. Is there a need for other object types to be dynamically passed down to the Processor?

Ideally any field can be specified at runtime. In the end, all objects and Classes we accept get serialized to a dict/JSON so we can handle this. It *is* possible to specify the value of any key in Parameters using JSONPath. 

Both proposed solutions require the stepfunctions SDK to maintain a mapping of each argument's location in the final CreateProcessingJob payload. I'm not sure we can derive this from the AWS SDK. Writing it by hand is more prone to mistakes and may lag behind if SageMaker adds new parameters to their APIs.

Here's a simpler solution with two parts that's more or less future-proof:

### 1. Each constructor argument we already accept today that is a primitive type can also be a `Placeholder`
We can hand-roll the Parameters substitution similar to #142 . We can be selective about which fields we want to support here.

### 2. Allow customers to use `parameters` to specify a dict containing static or placeholder values. 
In each SageMaker Task constructor, we accept `parameters` argument which is a `dict` containing static or `Placeholder` values. At the end of the constructor, we merge the autogenerated parameters from SageMaker classes and other constructor args with `parameters`. Adding this bit lets the customer use Placholders or static values for any field, including nested fields in SageMaker Python classes.

All non-SageMaker Tasks already accept a `parameters` arguemnt which automatically substitutes the keys for `Placeholders`. For SageMaker steps, [we override this](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L537).

wong-a (changes requested):
> Currently, only dicts, str and number objects were made placeholder compatible to facilitate testing purposes. Is there a need for other object types to be dynamically passed down to the Processor?

Ideally any field can be specified at runtime. In the end, all objects and Classes we accept get serialized to a dict/JSON so we can handle this. It *is* possible to specify the value of any key in Parameters using JSONPath. 

Both proposed solutions require the stepfunctions SDK to maintain a mapping of each argument's location in the final CreateProcessingJob payload. I'm not sure we can derive this from the AWS SDK. Writing it by hand is more prone to mistakes and may lag behind if SageMaker adds new parameters to their APIs.

Here's a simpler solution with two parts that's more or less future-proof:

### 1. Each constructor argument we already accept today that is a primitive type can also be a `Placeholder`
We can hand-roll the Parameters substitution similar to #142 . We can be selective about which fields we want to support here.

### 2. Allow customers to use `parameters` to specify a dict containing static or placeholder values. 
In each SageMaker Task constructor, we accept `parameters` argument which is a `dict` containing static or `Placeholder` values. At the end of the constructor, we merge the autogenerated parameters from SageMaker classes and other constructor args with `parameters`. Adding this bit lets the customer use Placholders or static values for any field, including nested fields in SageMaker Python classes.

All non-SageMaker Tasks already accept a `parameters` arguemnt which automatically substitutes the keys for `Placeholders`. For SageMaker steps, [we override this](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L537).

src/stepfunctions/steps/constants.py
ca-nguyen: I had the same thought - doing this by hand can introduce errors and is not easily maintainable.

Another option would be, for each SageMaker property, to call the function that adds it to the Parameters in the Sagemaker code instead of manually  getting the path from the [API docs](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html)
That would mean:
- Each property will call a different function (if existing) in sagemaker in order to add it to Parameters
- Some properties will not have an existing function in sagemaker that adds them to Parameters and we will have to do it by hand using the `placeholder_paths`

ca-nguyen:
Thank you both for your review!
@wong-a 's solution is less prone to errors and removes the need to map each arg to CreateProcessingJob request location.

This makes for a simpler and more effective solution and makes it the customer's responsibility to use a `parameters` structure that is up to date with SageMaker documents.

src/stepfunctions/steps/constants.py
ca-nguyen: Read @wong-a 's proposed solution after posting the previous comment - will go with that since it removes the need to map each args with `placeholder_paths`

src/stepfunctions/steps/constants.py
ca-nguyen: They are when passed to parameters
The placeholders are all replaced in parameters(included nested values) [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L51-#L65)

src/stepfunctions/steps/constants.py
ca-nguyen: 1: I took the location to the CreateProcessingJob request from there to save in placeholder_paths for each arg
2: all can

tests/integ/test_sagemaker_steps.py
```diff
@@ -384,27 +383,41 @@ def test_processing_step_with_placeholders(sklearn_processor_fixture, sagemaker_
 
     # Build workflow definition
     execution_input = ExecutionInput(schema={
-        Field.ImageUri.value: str,
-        Field.InstanceCount.value: int,
-        Field.Entrypoint.value: str,
-        Field.Role.value: str,
-        Field.VolumeSizeInGB.value: int,
-        Field.MaxRuntimeInSeconds.value: int
+        'image_uri': str,
```
ca-nguyen, line 386: Since we're only using these values for test purposes, using the direct string values for better code readability

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
...
+    """
+    Merges first and second dictionaries into the first one.
+    Values in the first dict are updated with the values of the second one.
+    """
+    if all(isinstance(d, dict) for d in [first, second]):
+        for key, value in second.items():
+            if key in first:
+                if isinstance(first[key], dict) and isinstance(second[key], dict):
+                    merge_dicts(first[key], second[key], first_name, second_name)
+                elif first[key] is value:
```
wong-a, line 61: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The `is` operator performs an identity comparison that uses the built-in `is()` method to test if the compared items are the same object or  not. It should be used with only singletons, such as `None`. We recommend that you use the `==` or `!=` equality operators to determine if two objects are equal.

[Learn more](https://docs.python.org/3/reference/datamodel.html#object.eq)

ca-nguyen: Good point! Will make the change to use `==` in the next commit

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
             return cur_partition
 
     return cur_partition
+
+
+def merge_dicts(first, second, first_name, second_name):
```
ca-nguyen, line 51: This could also be used to merge the [hyperparameters](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L175-#L180) in TrainingStep - will make the changes in another PR

src/stepfunctions/steps/sagemaker.py
```diff
@@ -518,22 +526,26 @@ def __init__(self, state_id, processor, job_name, inputs=None, outputs=None, exp
                                                                        SageMakerApi.CreateProcessingJob)
 
         if isinstance(job_name, str):
-            parameters = processing_config(processor=processor, inputs=inputs, outputs=outputs, container_arguments=container_arguments, container_entrypoint=container_entrypoint, kms_key_id=kms_key_id, job_name=job_name)
+            processing_parameters = processing_config(processor=processor, inputs=inputs, outputs=outputs, container_arguments=container_arguments, container_entrypoint=container_entrypoint, kms_key_id=kms_key_id, job_name=job_name)
         else:
-            parameters = processing_config(processor=processor, inputs=inputs, outputs=outputs, container_arguments=container_arguments, container_entrypoint=container_entrypoint, kms_key_id=kms_key_id)
+            processing_parameters = processing_config(processor=processor, inputs=inputs, outputs=outputs, container_arguments=container_arguments, container_entrypoint=container_entrypoint, kms_key_id=kms_key_id)
 
         if isinstance(job_name, Placeholder):
```
wong-a, line 533: thought: I was surprised we don't need to something like like this for `container_arguments`, `kms_key_id`, and `container_entrypoint` since [`processing_config` accepts `[str]` or `str` for these fields](https://github.com/aws/sagemaker-python-sdk/blob/master/src/sagemaker/workflow/airflow.py#L1108-L1110). However digging into the implementation, the values are just assigned to the corresponding CreateProcessingJob API field without any type checks.

https://github.com/aws/sagemaker-python-sdk/blob/467390f79e84310e0f987b3c19c095941b6925bd/src/sagemaker/processing.py#L961,L963

https://github.com/aws/sagemaker-python-sdk/blob/467390f79e84310e0f987b3c19c095941b6925bd/src/sagemaker/processing.py#L981-L982

src/stepfunctions/steps/sagemaker.py
wong-a: This is docstring is out of date with the new implementation

src/stepfunctions/steps/utils.py
wong-a: suggestion: First and second don't describe the side effects and which dict gets merged into what. Borrowing from [JavaScript's `Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#syntax):

```suggestion
def merge_dicts(target, source):
```

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
...
+    """
+    Merges first and second dictionaries into the first one.
```
wong-a, line 53: Please document the args for this function in the docstring

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
             return cur_partition
 
     return cur_partition
+
+
+def merge_dicts(first, second, first_name, second_name):
+    """
```
wong-a, line 52: Please add some unit tests for this function

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
...
+    Merges first and second dictionaries into the first one.
+    Values in the first dict are updated with the values of the second one.
+    """
+    if all(isinstance(d, dict) for d in [first, second]):
```
wong-a, line 56: For only two items, it's simpler as
```suggestion
    if isinstance(first, dict) and isinstance(second, dict):   
```

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,24 @@ def get_aws_partition():
...
+        for key, value in second.items():
+            if key in first:
+                if isinstance(first[key], dict) and isinstance(second[key], dict):
+                    merge_dicts(first[key], second[key], first_name, second_name)
+                elif first[key] == value:
+                    pass
+                else:
+                    logger.info(
+                        f"{first_name} property: <{key}> with value: <{first[key]}>"
+                        f" will be overwritten with value provided in {second_name} : <{value}>")
```
wong-a, line 66: question: Do we think this is useful? If not, can just use Python's built-in [`dict.update`](https://docs.python.org/3/library/stdtypes.html#dict.update)

src/stepfunctions/steps/sagemaker.py
```diff
@@ -477,9 +478,15 @@ class ProcessingStep(Task):
...
+    in the processing job (compatible with Placeholders): role, image_uri, instance_count, instance_type,
+    volume_size_in_gb, volume_kms_key, output_kms_key
     """
 
-    def __init__(self, state_id, processor, job_name, inputs=None, outputs=None, experiment_config=None, container_arguments=None, container_entrypoint=None, kms_key_id=None, wait_for_completion=True, tags=None, **kwargs):
+    def __init__(self, state_id, processor, job_name, inputs=None, outputs=None, experiment_config=None,
+                 container_arguments=None, container_entrypoint=None, kms_key_id=None, wait_for_completion=True,
+                 tags=None, **kwargs):
         """
         Args:
```
wong-a, line 491: Now that we accept `parameters`, document it here and describe what it is used for and how it relates to the rest of the args

wong-a (changes requested):
Looking close to finished with the new solution. Provided some minor comments for documentation and readability.

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Had the same thought! :)

src/stepfunctions/steps/utils.py
ca-nguyen: The built in update() does not take into account nested dictionary values - for ex:
```
d1 = {'a': {'aa': 1, 'bb': 2, 'c': 3}}
d2 = {'a': {'bb': 1}}

d1.update(d2)
print(d1)
```
Will have followint output: `{'a': {'bb': 1}}`

src/stepfunctions/steps/utils.py
ca-nguyen: Will add them with the next commit

src/stepfunctions/steps/utils.py
ca-nguyen: Will include args doc in the next commit

src/stepfunctions/steps/utils.py
ca-nguyen: Initially added them to facilitate troubleshooting, but I'm open to remove the logs if we deem them not useful enough or too noisy

wong-a: If the expected behaviour is well documented it seems unnecessary. Since the method only exists for logging, if we get rid of it there's less code to maintain. What do you think, @shivlaks?

src/stepfunctions/steps/utils.py
ca-nguyen: I was debating between both - so i'm good with the way you suggested! 
The changes will be included in the next commit

src/stepfunctions/steps/utils.py
wong-a: > The built in update() does not take into account nested dictionary values

Missed this comment. Since we need a deep merge, dict.update is not going to work here

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Updated the `parameters` doc [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/155/files#diff-bbe5592768d985044b425a113b47ee71849979921c0109dfa399e57e88f3bddbR506)

src/stepfunctions/steps/sagemaker.py
```diff
@@ -491,15 +494,19 @@ def __init__(self, state_id, processor, job_name, inputs=None, outputs=None, exp
...
-            tags (list[dict], optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
+            tags (list[dict] or Placeholder, optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
+            parameters(dict, optional): The value of this field becomes the request for the
+                `CreateProcessingJob<https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html>`_ created by the processing step.
+                All parameters fields are compatible with `Placeholders<https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=placeholder#stepfunctions.inputs.Placeholder>`_.
+                Any value defined in the parameters argument will overwrite the ones defined in the other arguments, including properties that were previously defined in the processor.
```
wong-a, line 509: > The value of this field becomes the request

Not 100% true, since these are merged with the values specified by other args.

```suggestion
            parameters(dict, optional): The value of this field is merged with other arguments to become the request payload for SageMaker `CreateProcessingJob<https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_CreateProcessingJob.html>`_. 
                You can use `parameters` to override the value provided by other arguments and specify any field's value dynamically using `Placeholders<https://aws-step-functions-data-science-sdk.readthedocs.io/en/stable/placeholders.html?highlight=placeholder#stepfunctions.inputs.Placeholder>`_.

```

src/stepfunctions/steps/sagemaker.py
```diff
@@ -518,22 +525,26 @@ def __init__(self, state_id, processor, job_name, inputs=None, outputs=None, exp
...
-        
-        kwargs[Field.Parameters.value] = parameters
+            processing_parameters['Tags'] = tags if isinstance(tags, Placeholder) else tags_dict_to_kv_list(tags)
+
+        if 'S3Operations' in processing_parameters:
+            del processing_parameters['S3Operations']
+
+        if Field.Parameters.value in kwargs and isinstance(kwargs[Field.Parameters.value], dict):
+            # Update processing_parameters with input parameters
+            merge_dicts(processing_parameters, kwargs[Field.Parameters.value], "Processing Parameters",
```
wong-a, line 546: I don't think "Processing Parameters" and "Input Parameters" will make much sense in the logs from the customer's POV. There's no argument to the constructor with either name. Would rather make the log message generic instead.

tests/unit/test_steps_utils.py
```diff
@@ -51,3 +51,10 @@ def test_arn_builder_sagemaker_wait_completion():
                                       IntegrationPattern.WaitForCompletion)
     assert arn == "arn:aws:states:::sagemaker:createTrainingJob.sync"
 
+
+def test_merge_dicts():
+    d1 = {'a': {'aa': 1, 'bb': 2, 'cc': 3}, 'b': 1}
+    d2 = {'a': {'bb': {'aaa': 1, 'bbb': 2}}, 'b': 2, 'c': 3}
+
+    merge_dicts(d1, d2, 'd1', 'd2')
+    assert d1 == {'a': {'aa': 1, 'bb': {'aaa': 1, 'bbb': 2}, 'cc': 3}, 'b': 2, 'c': 3}
```
wong-a, line 60: nit: indent the dicts to make it easier to read

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Good catch!

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agree - this could confuse the customer even further. Will change to generic log message in the next commit

tests/unit/test_steps_utils.py
ca-nguyen: Always for improving code readability - changes will be included in the next commit

wong-a (approved):

tests/integ/test_sagemaker_steps.py
```diff
@@ -352,3 +353,100 @@ def test_processing_step(sklearn_processor_fixture, sagemaker_session, sfn_clien
...
+                                     inputs=inputs,
+                                     outputs=outputs,
+                                     container_arguments=execution_input['container_arguments'],
+                                     container_entrypoint=execution_input['entrypoint'],
+                                     parameters=parameters
+                                     )
+    workflow_graph = Chain([processing_step])
+
+    with timeout(minutes=DEFAULT_TIMEOUT_MINUTES):
+        # Create workflow and check definition
```
shivlaks, line 425: nit: unnecessary comment as the method name expresses this in snake case

shivlaks (approved):
I like this approach as it's much closer to what we do in the CDK for property bags vs. maintaining hand-rolled properties, which has proven to be untenable in the past. In this case, it was also becoming unwieldy due to the number of properties in these APIs.

had a couple nits and a question or two, but overall it looks good!

The summary indicates that this closes #85 but that one doesn't seem to be an issue for processing step. please amend if needed before you merge.

src/stepfunctions/steps/utils.py
```diff
@@ -45,3 +46,28 @@ def get_aws_partition():
             return cur_partition
 
     return cur_partition
+
+
+def merge_dicts(target, source):
```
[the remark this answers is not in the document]
ca-nguyen: This was implemented having [dict.update()](https://python-reference.readthedocs.io/en/latest/docs/dict/update.html) function in mind, where it is possible to update a dict with another. In our case, we are merging nested dicts as well. 
Mutable objects are all passed by reference in Python and the description explains the function behaviour, so I think it makes sense to leave it as is - what do you think? :)

src/stepfunctions/steps/sagemaker.py
```diff
@@ -30,6 +30,7 @@
 
 SAGEMAKER_SERVICE_NAME = "sagemaker"
 
+
```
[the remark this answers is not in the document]
ca-nguyen: 🙌🙌🙌

src/stepfunctions/exceptions.py
```diff
@@ -22,4 +22,5 @@ class MissingRequiredParameter(Exception):
 
 
 class DuplicateStatesInChain(Exception):
-    pass
\ No newline at end of file
+    pass
+
```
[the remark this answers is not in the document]
ca-nguyen: Good point - will remove it from this PR

tests/integ/test_sagemaker_steps.py
```diff
@@ -352,3 +353,100 @@ def test_processing_step(sklearn_processor_fixture, sagemaker_session, sfn_clien
         # Cleanup
         state_machine_delete_wait(sfn_client, workflow.state_machine_arn)
         # End of Cleanup
+
+
+def test_processing_step_with_placeholders(sklearn_processor_fixture, sagemaker_session, sfn_client, sfn_role_arn,
+                                           sagemaker_role_arn):
+    region = boto3.session.Session().region_name
+    input_data = 's3://sagemaker-sample-data-{}/processing/census/census-income.csv'.format(region)
```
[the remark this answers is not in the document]
ca-nguyen: I agree that using `fstring` is more readable and efficient. `format` was used for all other tests so i kept it for consistency.
Will change it for this added test and perhaps we can make the change for the rest of the file in a separate PR

tests/integ/test_sagemaker_steps.py
```diff
@@ -352,3 +353,100 @@ def test_processing_step(sklearn_processor_fixture, sagemaker_session, sfn_clien
...
+
+    input_s3 = sagemaker_session.upload_data(
+        path=os.path.join(DATA_DIR, 'sklearn_processing'),
+        bucket=sagemaker_session.default_bucket(),
+        key_prefix='integ-test-data/sklearn_processing/code'
+    )
+
+    output_s3 = 's3://' + sagemaker_session.default_bucket() + '/integ-test-data/sklearn_processing'
```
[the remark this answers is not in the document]
ca-nguyen: Agreed - using `fstring`would be more readable and efficient. 

Same comment: `format` was used for all other tests so i kept it for consistency.
Will change it for this added test and perhaps we can make the change for the rest of the file in a separate PR

tests/integ/test_sagemaker_steps.py
ca-nguyen: Yes  - will remove it in the next commit!

tests/integ/test_sagemaker_steps.py
ca-nguyen: Agreed- will be removed with the next commit

ca-nguyen:
> The summary indicates that this closes #85 but that one doesn't seem to be an issue for processing step. please amend if needed before you merge.

You are right! will remove it from the linked issues!

shivlaks (approved):

ca-nguyen:
Received 2 ship-its - merging PR

wong-a (approved):

Issue #101, opened 2020-10-21
Feature Request: Support placeholders for Map state


mochacafe:
Currently I am not able to pass execution input to a Map state. If I try to pass execution input to map state it throws 'Object of type 'ExecutionInput' is not JSON serializable' . So is it possible to add palceholder support to Map state.

```
input_data_check_map_step= steps.states.Map(state_id='input_data_check_map',
                                           iterator=input_data_check_lambda_task,
                                            input_path=execution_input,
                                           items_path=execution_input['input_data_dirs'],
                                           parameters={'input_dir':'$.Map.Item.Value','fcst_date':"$.fcst_date"},
                                           max_concurrency=5,
                                           comment='Run input data checks in dynamic parllel state',
                                           result_path='$.InputDataCheckRes')
```

yoodan93:
Step Functions recently added support for execution input in InputPath and ItemsPath for Map state! https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html

The python SDK still does not support this, but we have plans of keeping the python SDK up to date to current Amazon States Language specs.

wong-a:
"Support placeholders for Map state" is a broad topic - let's clarify what this means and enumerate the specific use cases and feature gaps.

In @mochacafe's code example there are two places where placeholders are not trying to be used but not supported:

### 1. InputPath should accept Placeholders (and also OutputPath and ResultPath)
```python
input_path=execution_input
```

The generated ASL should be the following. 
```json
"InputPath": "$$.Execution.Input"
```

Note that unlike Placeholders in `Parameters` (the only place they are supported today) the original key **should not be** suffixed by ".$". All the "Path" ASL fields are just JSONPath strings. 

This allows you to specify InputPath without writing a literal JSONPath string by hand. Today *you can use the context object (or any JSONPath)* in input_path but you need to pass a JSONPath string (e.g. `input_path='$$.Execution.Input''`) like how `result_path` is set in the example. The Placeholder class is just a pythonic way to construct the dynamic references without writing JSONPath strings by hand (plus some optional static validation if schemas are used).

This applies to all states that support the InputPath field, not just Map states. OutputPath and ResultPath can have the same functionality too. One caveat is that ResultPath shouldn't accept ExecutionInput, since the context object is non-writable.

### 2. ItemsPath should accept Placeholders 

Similar to above, make it easier to construct the JSONPath using Placeholders. The workaround today is to just write the string:

```
[...]

PR #102, opened 2020-10-29
Add support for Amazon States Language "ResultSelector" in Task, Map …

yoodan93:
Add ResultSelector support for Task, Map and Parallel states.

*Issue #, if available:*
https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/95

---
#### Description of changes

Adding new Amazon States Language payload template "ResultSelector" https://states-language.net/#payload-template
This change allows user to define a `result_selector` which is used to define the effective result of a state.

#### Testing
- Added unit test for changes
- Manual test:

```
# Create a simple workflow containing a task state with `result_selector`
from stepfunctions.inputs import StepResult

step_result = StepResult()
step_result_placeholder = {
    'Output': step_result['output']
}

state_with_result_selector = Task(
    state_id='StartState',
    parameters={'Key': 'Value'},
    result_selector=step_result_placeholder,
    resource=<resource_arn>
)

workflow = Workflow(
    name="ResultSelector_Example",
    definition=state_with_result_selector,
    role=<workflow_execution_role>
)

# Execute the workflow and confirm it has completed successfully
workflow.execute()
```
Workflow definition: 
```
{
    "StartAt": "StartState",
    "States": {
        "StartState": {
            "Parameters": {
                "Key": "Value"
            },
            "ResultSelector": {
                "Output.$": "$['output']"
            },
            "Resource": <resource_arn>,
            "Type": "Task",
            "End": true
        }
    }
}
```

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #158, opened 2021-09-03
feat: Support placeholders for Map state

ca-nguyen:
*Issue #, if available:* #101

*Description of changes:*
Currently, it is not possible to use Placeholder with Map state to set the values dynamically outside of properties passed from `parameters`.
With this change, it will be possible to do so with following args: 
- items_path 
- comment 
- input_path 
- result_path
- output_path

*Open question*
Making the [common fields](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/fields.py#L20-L25) (comment, input_path, result_path, output_path) compatible with placeholders will allow customers to use Placeholders for those fields in other States too.
The documentation will need to be updated for those states - should this doc change be included in this PR as well? 
Since those are minor changes, I am more inclined to include them in here, but am open to do either to facilitate the review process.

### DIFF
--- a/src/stepfunctions/steps/states.py
+++ b/src/stepfunctions/steps/states.py
@@ -73,10 +73,36 @@ class Block(object):
                 k = to_pascalcase(k)
                 if k == to_pascalcase(Field.Parameters.value):
                     result[k] = self._replace_placeholders(v)
+                elif self._is_placeholder_compatible(k):
+                    if isinstance(v, Placeholder):
+                        modified_key = f"{k}.$"
+                        result[modified_key] = v.to_jsonpath()
+                    else:
+                        result[k] = v
                 else:
                     result[k] = v
 
         return result
+
+    @staticmethod
+    def _is_placeholder_compatible(field):
+        """
+        Check if the field is placeholder compatible
+
+        Args:
+            field: Field against which to verify placeholder compatibility
+        """
+        return field in [
+            # Common fields
+            to_pascalcase(Field.Comment.value),
+            to_pascalcase(Field.InputPath.value),
+            to_pascalcase(Field.OutputPath.value),
+            to_pascalcase(Field.ResultPath.value),
+
+            # Map
+            to_pascalcase(Field.ItemsPath.value),
+            to_pascalcase(Field.MaxConcurrency.value),
+        ]
 
     def to_json(self, pretty=False):
         """Serialize to a JSON formatted string.
@@ -541,13 +567,13 @@ class Map(State):
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
-            items_path (str, optional): Path in the input for items to iterate over. (default: '$')
-            max_concurrency (int, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
-            comment (str, optional): Human-readable comment or description. (default: None)
-            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            items_path (str or Placeholder, optional): Path in the input for items to iterate over. (default: '$')
+            max_concurrency (int or Placeholder, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
+            comment (str or Placeholder, optional): Human-readable comment or description. (default: None)
+            input_path (str or Placeholder, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
             parameters (dict, optional): The value of this field becomes the effective input for the state.
-            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
-            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+            result_path (str or Placeholder, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str or Placeholder, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
         """
         super(Map, self).__init__(state_id, 'Map', **kwargs)
 
--- a/tests/unit/test_placeholders_with_steps.py
+++ b/tests/unit/test_placeholders_with_steps.py
@@ -214,6 +214,69 @@ def test_map_state_with_placeholders():
     result = Graph(workflow_definition).to_dict()
     assert result == expected_repr
 
+
+def test_map_state_with_placeholders():
+    workflow_input = ExecutionInput(schema={
+        'comment': str,
+        'input_path': str,
+        'output_path': str,
+        'result_path': str,
+        'items_path': str,
+        'max_concurrency': int,
+        'ParamB': str
+    })
+
+    map_state = Map(
+        'MapState01',
+        comment=workflow_input['input_path'],
+        input_path=workflow_input['input_path'],
+        output_path=workflow_input['output_path'],
+        result_path=workflow_input['result_path'],
+        items_path=workflow_input['result_path'],
+        max_concurrency=workflow_input['max_concurrency']
+    )
+    iterator_state = Pass(
+        'TrainIterator',
+        parameters={
+            'ParamA': map_state.output()['X']["Y"],
+            'ParamB': workflow_input['ParamB']
+    })
+
+    map_state.attach_iterator(iterator_state)
+    workflow_definition = Chain([map_state])
+
+    expected_repr = {
+        "StartAt": "MapState01",
+        "States": {
+            "MapState01": {
+                "Type": "Map",
+                "End": True,
+                "Comment.$": "$$.Execution.Input['input_path']",
+                "InputPath.$": "$$.Execution.Input['input_path']",
+                "ItemsPath.$": "$$.Execution.Input['result_path']",
+                "Iterator": {
+                    "StartAt": "TrainIterator",
+                    "States": {
+                        "TrainIterator": {
+                            "Parameters": {
+                                "ParamA.$": "$['X']['Y']",
+                                "ParamB.$": "$$.Execution.Input['ParamB']"
+                            },
+                            "Type": "Pass",
+                            "End": True
+                        }
+                    }
+                },
+                "MaxConcurrency.$": "$$.Execution.Input['max_concurrency']",
+                "OutputPath.$": "$$.Execution.Input['output_path']",
+                "ResultPath.$": "$$.Execution.Input['result_path']",
+            }
+        }
+    }
+
+    result = Graph(workflow_definition).to_dict()
+    assert result == expected_repr
+
 def test_parallel_state_with_placeholders():
     workflow_input = ExecutionInput()
 


src/stepfunctions/steps/states.py
```diff
@@ -541,13 +567,13 @@ def __init__(self, state_id, **kwargs):
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
-            items_path (str, optional): Path in the input for items to iterate over. (default: '$')
-            max_concurrency (int, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
-            comment (str, optional): Human-readable comment or description. (default: None)
-            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            items_path (str or Placeholder, optional): Path in the input for items to iterate over. (default: '$')
+            max_concurrency (int or Placeholder, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
+            comment (str or Placeholder, optional): Human-readable comment or description. (default: None)
```
wong-a, line 572: Comments are not compatible with placeholders. It's just a static string

src/stepfunctions/steps/states.py
```diff
@@ -541,13 +567,13 @@ def __init__(self, state_id, **kwargs):
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
-            items_path (str, optional): Path in the input for items to iterate over. (default: '$')
-            max_concurrency (int, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
-            comment (str, optional): Human-readable comment or description. (default: None)
-            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            items_path (str or Placeholder, optional): Path in the input for items to iterate over. (default: '$')
+            max_concurrency (int or Placeholder, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
```
wong-a, line 571: I don't think this one is either

src/stepfunctions/steps/states.py
ca-nguyen: Removed it in the latest commit

src/stepfunctions/steps/states.py
ca-nguyen: You are right - Removed it in the latest commit

tests/unit/test_placeholders.py
```diff
@@ -178,4 +178,170 @@ def check_immutable(placeholder):
             for k, v in placeholder.store.items():
                 return check_immutable(v)
     else:
-        return False
\ No newline at end of file
+        return False
+
+
+def test_map_item_value_creation_with_subscript_operator():
```
ca-nguyen, line 184: Tests will be parametrized once [ResultSelector PR](https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/102) is merged

src/stepfunctions/inputs/placeholders.py
```diff
@@ -291,3 +291,50 @@ def _create_variable(self, name, parent, type=None):
...
+        else:
+            return MapItemValue(name=name, parent=parent)
+
+
+class MapItemIndex(Placeholder):
+    """
+        Top-level class for map item index placeholders.
+    """
+
+    def __init__(self, schema=None, **kwargs):
```
shivlaks, line 323: question: is there a test that uses / sets this Schema

tests/unit/test_placeholders_with_steps.py
```diff
@@ -214,6 +214,64 @@ def test_map_state_with_placeholders():
...
+        }
+    )
+    iterator_state = Pass(
+        'TrainIterator',
+        parameters={
+            'ParamA': map_state.output()['X']["Y"]
+        }
+    )
+
+    map_state.attach_iterator(iterator_state)
```
shivlaks, line 241: question: what happens with this? - does it become part of the workflow definition somewhere? I thought the `Iterator` and `ItemsPath` properties were required for `Map` states

tests/unit/test_placeholders_with_steps.py
```diff
@@ -214,6 +214,64 @@ def test_map_state_with_placeholders():
...
+        'name': str,
+        'age': str
+    })
+
+    map_item_index = MapItemIndex()
+
+    map_state = Map(
+        'MapState01',
+        parameters={
+            "MapIndex": map_item_index,
```
shivlaks, line 229: question: is using multiple indexes valid? - seems we don't preclude the ability to do that, but would that generate invalid ASL? - does it make sense to protect users against it or is it overkill right now?

shivlaks (changes requested):
thanks for putting this PR together! left some feedback, let me know if you have any questions. 

some higher level thoughts:

* I think we should also update the [docs](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/ab08aad7e80112dd9ca9dd02863765b1c4770ca0/doc/placeholders.rst) related to placeholders. Might make sense to re-purpose the example in your commit body to illustrate its usage.
* does it make sense to add an integration test that exercises placeholders? as we expand the use cases and scenarios we support, I'm thinking it would be useful to have some basic integ tests as unit tests are more brittle.

tests/unit/test_placeholders_with_steps.py
wong-a: ItemsPath is optional. Similar to InputPath and OutputPath, the default is `$`.

src/stepfunctions/steps/states.py
```diff
@@ -541,7 +541,7 @@ def __init__(self, state_id, **kwargs):
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
-            items_path (str, optional): Path in the input for items to iterate over. (default: '$')
+            items_path (str or Placeholder, optional): Path in the input for items to iterate over. (default: '$')
```
wong-a, line 544: Needs a unit test. Do the existing fields: `input_path`, `result_path`, and `output_path` even support Placeholders today? ItemsPath is similar to those in that it's key is never suffixed by `.$` like Parameter keys referencing a Path. Its value is always treated a JSONPath, never a regular string. 

In addition to `items_path`, https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/101 is also trying to use `ExecutionInput` for `input_path`.

tests/unit/test_placeholders_with_steps.py
wong-a: Parameters can be anything you want. You can reference `$$.Map.Item.Index` as multiple times for different fields, but the value is always the same within the iteration.

wong-a (changes requested):

src/stepfunctions/inputs/placeholders.py
```diff
@@ -291,3 +291,50 @@ def _create_variable(self, name, parent, type=None):
...
+        super(MapItemValue, self).__init__(schema, **kwargs)
+        self.json_str_template = '$$.Map.Item.Value{}'
+
+    def _create_variable(self, name, parent, type=None):
+        """
+            Creates a placeholder variable for Map Item Value.
+            A placeholder variable can only be created if the collection is not immutable due to a pre-specified schema.
+        """
+        if self.immutable:
+            raise ValueError("Placeholder variable does not conform to schema set for the placeholder collection.")
```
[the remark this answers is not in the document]
ca-nguyen: Agreed - will add the provided schema description in the error log

tests/unit/test_placeholders_with_steps.py
```diff
@@ -214,6 +214,64 @@ def test_map_state_with_placeholders():
...
     assert result == expected_repr
 
+
+def test_map_state_with_placeholders():
+    map_item_value = MapItemValue(schema={
+        'name': str,
+        'age': str
+    })
+
+    map_item_index = MapItemIndex()
```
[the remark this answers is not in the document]
ca-nguyen: Agreed - will make the change in the next commit

src/stepfunctions/inputs/placeholders.py
ca-nguyen: Yes, there is - test_map_item_value_with_schema() in tests/unit/test_placeholders.py uses the schema
We validate the schema and the exceptions that are thrown if the schema is not respected
I can add an assert for success as well

src/stepfunctions/steps/states.py
ca-nguyen: You're right - i forgot to remove the Placeholder mention from an earlier commit. 
We don't currently support Placeholders for `items_path`, `input_path`, `result_path`, and `output_path`.
I can add support in the next commit

ca-nguyen: I can add support in my next commit

ca-nguyen: I tested in the AWSConsole and currently, we don't support Placeholders for `items_path`, `input_path`, `result_path` and `output_path`

Tested with a state machine based on [Map State Example](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html)
```
{
  "StartAt": "MapExample01",
  "States": {
    "MapExample01": {
      "InputPath": "$$.Execution.Input['input_path']",
      "ResultPath": "$.detail.shipping",  # Note: State machine would fail to create with execution input
      "ItemsPath": "$$.Execution.Input['items_path']",
      "OutputPath": "$$.Execution.Input['output_path']",
      "Type": "Map",
      "End": true,
      "Iterator": {
        "StartAt": "Validate",
        "States": {
          "Validate": {
            "Resource": <resource_arn>,
            "Type": "Task",
            "End": true
          }
        }
      }
    }
  }
}
```

With execution input:

```
{
  "ship-date": "2016-03-14T01:59:00Z",
  "detail": {
    "delivery-partner": "UQS",
    "shipped": [
      {
        "prod": "R31",
        "dest-code": 9511,
        "quantity": 1344
      },
      {
        "prod": "S39",
        "dest-code": 9511,
        "quantity": 40
      }
    ]
  },
  "input_path": "$.detail",
  "items_path": "$.shipped",
  "output_path": "$.detail.shipped"
}
```
Execution would fail as paths could not be applied to execution input

I will remove the Placeholder mention from the doc

src/stepfunctions/inputs/placeholders.py
ca-nguyen: Just read this again and realizing this is for `MapItemIndex` and not `MapItemValue`
The `MapItemIndex` should not have a schema as it can only return the index of the map iteration - I will make the change in the next commit

src/stepfunctions/steps/states.py
wong-a: That's not true. InputPath, ItemsPath, OutputPath, and ResultPath must be JSONPath values. The exception you got during execution is because the payload between input/output processing steps didn't have the right data.

```
Raw input -> InputPath  -> ItemsPath  ->  Parameters  -> ResultPath  -> Output Path
```

The example you used doesn't work because the value eferenced in the context object is a JSONPath string instead of an array. For ItemsPath, the dereferenced runtime value is the literal string "$.shipped". Which is not an array. Step Functions doesn't treat the value as another JSONPath to interpolate again. You need to set `"InputPath": "$.shipped"` and `"ItemsPath": "$.items"` in the state machine definition or reference them from the context objext ($$.Execution.Input.shipped).

"Placeholder" is a data science SDK concept that doesn't exist in ASL that makes it easier to reference dynamic variables in a pythonic way without writing JSONPath.

ca-nguyen: Thanks for your input!
I confirm that it does work to use the reference directly

```
{
  "StartAt": "MapExample01",
  "States": {
    "MapExample01": {
      "InputPath": "$$.Execution.Input['input_path']",
      "ResultPath": "$.detail.shipped",
      "ItemsPath": "$.shipped",
      "OutputPath": "$.detail.shipped",
      "Type": "Map",
      "End": true,
      "Iterator": {
        "StartAt": "Validate",
        "States": {
          "Validate": {
            "Resource": <resource_arn>,
            "Type": "Task",
            "End": true
          }
        }
      }
    }
  }
}
```

with this execution input

```
{
  "ship-date": "2016-03-14T01:59:00Z",
  "detail": {
    "delivery-partner": "UQS"
  },
  "input_path": {
    "delivery-partner": "UQS",
    "shipped": [
      {
        "prod": "R31",
        "dest-code": 9511,
        "quantity": 1344
      },
      {
        "prod": "S39",
        "dest-code": 9511,
        "quantity": 40
      }
    ]
  }
}
```

ca-nguyen: Change to support `itemspath` and `inputpath` with placeholders will be added with the next commit

wong-a:
Clarified what "support placeholders for Map state" means in the initial issue comments: https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/101#issuecomment-917040833. So far, this PR only addresses the 3rd use case, which isn't actually what the requester was trying to do.

ca-nguyen:
>I think we should also update the docs related to placeholders. Might make sense to re-purpose the example in your commit body to illustrate its usage.

Agreed - will update the placeholder docs 

>does it make sense to add an integration test that exercises placeholders? as we expand the use cases and scenarios we support, I'm thinking it would be useful to have some basic integ tests as unit tests are more brittle.

Yes it does - will include one for Map state

wong-a:
@ca-nguyen This change doesn't just affect Map state. There's the 3 things I mentioned here: https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/101#issuecomment-917040833

InputPath and OutputPath are allowed in all state types except Fail. Please update the PR title, description, and tests accordingly.

src/stepfunctions/steps/states.py
```diff
@@ -73,11 +73,33 @@ def to_dict(self):
                 k = to_pascalcase(k)
                 if k == to_pascalcase(Field.Parameters.value):
                     result[k] = self._replace_placeholders(v)
+                elif self._is_placeholder_compatible(k):
+                    if isinstance(v, Placeholder):
+                        result[k] = v.to_jsonpath()
+                    else:
+                        result[k] = v
```
wong-a, line 80: I think this would be clearer without nested if statements. The final `else` is the same as the one already on line 81
```suggestion
                elif self._is_placeholder_compatible(k) and isinstance(v, Placeholder):
                    result[k] = v.to_jsonpath()

```

src/stepfunctions/steps/states.py
```diff
@@ -541,13 +563,13 @@ def __init__(self, state_id, **kwargs):
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
-            items_path (str, optional): Path in the input for items to iterate over. (default: '$')
+            items_path (str or Placeholder, optional): Path in the input for items to iterate over. (default: '$')
             max_concurrency (int, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
             comment (str, optional): Human-readable comment or description. (default: None)
-            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            input_path (str or Placeholder, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
```
wong-a, line 569: This applies to all states except Fail, so let's update the docstrings for those steps too and make sure they are tested.

src/stepfunctions/steps/states.py
```diff
@@ -541,13 +563,13 @@ def __init__(self, state_id, **kwargs):
...
             parameters (dict, optional): The value of this field becomes the effective input for the state.
             result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
```
wong-a, line 571: not-blocking: Why is result_path omitted?

tests/unit/test_placeholders_with_steps.py
```diff
@@ -176,8 +176,22 @@ def test_step_input_order_validation():
 
 def test_map_state_with_placeholders():
     workflow_input = ExecutionInput()
```
wong-a, line 178: note: Placeholders don't have to be `ExecutionInput` all the time. You can use `StepInput` too

wong-a (dismissed):

src/stepfunctions/steps/states.py
ca-nguyen: Good point - will update the docstrings in the next commit

tests/unit/test_placeholders_with_steps.py
ca-nguyen: I'll make sure to include both in next commit

src/stepfunctions/steps/states.py
ca-nguyen: The result_path was omitted as it only works with the $[], but i agree that it can be included in this PR as well, will add that in next commits

wong-a: What's `$[]`? You mean it doesn't work with the context object?

Let's keep it out for now. Just create a separate issue to set resultPath with Placeholders.

For the record: every question question I ask doesn't need to be implemented. You are free (and expected!) to push back if you find a reason. It's best to keep the scope of PRs small anyways

ca-nguyen: Yes, that is correct - it does not work with a context object
> For the record: every question question I ask doesn't need to be implemented. You are free (and expected!) to push back if you find a reason. It's best to keep the scope of PRs small anyways

ACK, I will keep that in mind moving forward

src/stepfunctions/steps/states.py
ca-nguyen: Indeed - this is clearer!

src/stepfunctions/steps/states.py
ca-nguyen: Updated docstrings and added tests in latest commit

ca-nguyen:
>InputPath and OutputPath are allowed in all state types except Fail. Please update the PR title, description, and tests accordingly.

Updated the PR title and description

tests/unit/test_placeholders_with_steps.py
ca-nguyen: The iterator is added to the Map State [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L554-L561) and will be added to the workflow definition when the Map state itself is added to the workflow definition

The [docstring](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L543) will need to be updated - it seems like `iterator` is required in the Map State constructor when in fact, it gets creates successfully without it. However, the workflow will fail to create if no iterator is set - we could add a validation [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L578) and raise an exception if the iterator is not set. At the moment, if fails [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/states.py#L735) with the following message
```
ValueError: Expected branch to be a State or a Chain, but got `None`
```

This can be done in another PR to avoid making the scope larger

Issue #152, opened 2021-07-13
TrainingStep's hyperparameters does not work on v2.2.0


ohbuchim:
I cannot set TrainingStep's 'hyperparameters' parameter using ExecutionInput on v2.2.0. v2.1.0 works well.

When I used v2.1.0 this code worked, and I was able to set hyperparameters when I executed workflow.
```
training_step = steps.TrainingStep(
    "SageMaker Training Step",
    estimator=sklearn,
    data={"train": sagemaker.TrainingInput(execution_input["PreprocessingOutputDataTrain"], content_type="text/csv")},
    job_name=execution_input["TrainingJobName"],
    hyperparameters=execution_input["TrainingParameters"],
    wait_for_completion=True,
)
```
When I used v2.2.0 I needed to add get_schema_as_dict() to ExecutionInput. After that, 'hyperparameters' was not updated by ExecutionInput. It seems that the estimator's 'hyperparameters' setting only is set (ExecutionInput is ignored).
```
training_step = steps.TrainingStep(
    "SageMaker Training Step",
    estimator=sklearn,
    data={"train": sagemaker.TrainingInput(execution_input["PreprocessingOutputDataTrain"], content_type="text/csv")},
    job_name=execution_input["TrainingJobName"],
    hyperparameters=execution_input["TrainingParameters"].get_schema_as_dict(),
    wait_for_completion=True,
)
```

shivlaks:
@ohbuchim - thanks for reporting the issue, it does seem to be a regression due to merging the hyperparameters set in the estimator along with the ones established in the constructor for `TrainingStep`

in `v2.1.0`

it would resolve to something like the following, but it would also get rid of any hyperparameters that were established in the estimator by doing so.

```
'HyperParameters.$': "$$.Execution.Input['TrainingParameters']"
```

does the workaround actually work though? I think it just discards the constructor parameter altogether.

To support placeholders, I'm thinking perhaps we should skip merging estimator established hyperparameters if a placeholder is provided. Note that this will result in us dropping any hyperparameters that were established in the estimator (as it was with `v2.1.0`.

what do you think?

ohbuchim:
Hi, thanks for your comments. I expected estimator's 'hyperparameters' and ExecutionInput's 'hyperparameters' would be merged.

[...]

PR #159, opened 2021-09-09
fix: Support placeholders for hyperparameters passed to TrainingStep

ca-nguyen:
*Issue #, if available:* #152

*Description of changes:*
A breaking change was introduced in V2.2.0 where it was no longer possible to use Placeholders as input for `hyperparameters` in TrainingStep.

With this solution, TrainingStep `hyperparameters` will be compatible with Placeholders with the same behaviour as in V2.1.0: if a Placeholder is provided as input to `hyperparameters` in TrainingStep constructor, the estimator hyperparameters are overwritten
 
A feature request to be able to merge the Placeholder hyperparameters was made in #152, which will be addressed in another PR.

### DIFF
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -72,7 +72,7 @@ class TrainingStep(Task):
             hyperparameters: Parameters used for training.
                 * (dict[str, str], optional) - Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
                     If there are duplicate entries, the value provided through this property will be used. (Default: Hyperparameters specified in the estimator.)
-                * (Placeholder, optional) - Hyperparameters supplied will overwrite the Hyperparameters specified in the estimator.
+                * (Placeholder, optional) - The TrainingStep will use the hyperparameters specified by the Placeholder's value instead of the hyperparameters specified in the estimator.
             mini_batch_size (int): Specify this argument only when estimator is a built-in estimator of an Amazon algorithm. For other estimators, batch size should be specified in the estimator.
             experiment_config (dict, optional): Specify the experiment config for the training. (Default: None)
             wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait for the training job to complete before proceeding to the next step in the workflow. Set to `False` if the Task state should submit the training job and proceed to the next step. (default: True)
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -69,7 +69,7 @@ class TrainingStep(Task):
                 * (list[sagemaker.amazon.amazon_estimator.RecordSet]) - A list of
                     :class:`sagemaker.amazon.amazon_estimator.RecordSet` objects,
                     where each instance is a different channel of training data.
-            hyperparameters (dict, optional): Parameters used for training.
+            hyperparameters (dict[str, str] or dict[str, Placeholder], optional): Parameters used for training.
                     Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
                     If there are duplicate entries, the value provided through this property will be used. (Default: Hyperparameters specified in the estimator.)
             mini_batch_size (int): Specify this argument only when estimator is a built-in estimator of an Amazon algorithm. For other estimators, batch size should be specified in the estimator.
@@ -127,9 +127,8 @@ class TrainingStep(Task):
             parameters['InputDataConfig'][0]['DataSource']['S3DataSource']['S3Uri.$'] = data_uri
 
         if hyperparameters is not None:
-            if not isinstance(hyperparameters, Placeholder):
-                if estimator.hyperparameters() is not None:
-                    hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
+            if estimator.hyperparameters() is not None:
+                hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
             parameters['HyperParameters'] = hyperparameters
 
         if experiment_config is not None:
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -127,8 +127,9 @@ class TrainingStep(Task):
             parameters['InputDataConfig'][0]['DataSource']['S3DataSource']['S3Uri.$'] = data_uri
 
         if hyperparameters is not None:
-            if estimator.hyperparameters() is not None:
-                hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
+            if not isinstance(hyperparameters, Placeholder):
+                if estimator.hyperparameters() is not None:
+                    hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
             parameters['HyperParameters'] = hyperparameters
 
         if experiment_config is not None:
@@ -173,12 +174,7 @@ class TrainingStep(Task):
             estimator_hyperparameters (dict): Hyperparameters specified in the estimator
         """
         merged_hyperparameters = estimator_hyperparameters.copy()
-        for key, value in training_step_hyperparameters.items():
-            if key in merged_hyperparameters:
-                logger.info(
-                    f"hyperparameter property: <{key}> with value: <{merged_hyperparameters[key]}> provided in the"
-                    f" estimator will be overwritten with value provided in constructor: <{value}>")
-            merged_hyperparameters[key] = value
+        merge_dicts(merged_hyperparameters, training_step_hyperparameters)
         return merged_hyperparameters
 
 class TransformStep(Task):
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -69,9 +69,10 @@ class TrainingStep(Task):
                 * (list[sagemaker.amazon.amazon_estimator.RecordSet]) - A list of
                     :class:`sagemaker.amazon.amazon_estimator.RecordSet` objects,
                     where each instance is a different channel of training data.
-            hyperparameters (dict[str, str] or dict[str, Placeholder], optional): Parameters used for training.
-                    Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
+            hyperparameters: Parameters used for training.
+                * (dict[str, str], optional) - Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
                     If there are duplicate entries, the value provided through this property will be used. (Default: Hyperparameters specified in the estimator.)
+                * (Placeholder, optional) - Hyperparameters supplied will overwrite the Hyperparameters specified in the estimator.
             mini_batch_size (int): Specify this argument only when estimator is a built-in estimator of an Amazon algorithm. For other estimators, batch size should be specified in the estimator.
             experiment_config (dict, optional): Specify the experiment config for the training. (Default: None)
             wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait for the training job to complete before proceeding to the next step in the workflow. Set to `False` if the Task state should submit the training job and proceed to the next step. (default: True)
@@ -127,8 +128,9 @@ class TrainingStep(Task):
             parameters['InputDataConfig'][0]['DataSource']['S3DataSource']['S3Uri.$'] = data_uri
 
         if hyperparameters is not None:
-            if estimator.hyperparameters() is not None:
-                hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
+            if not isinstance(hyperparameters, Placeholder):
+                if estimator.hyperparameters() is not None:
+                    hyperparameters = self.__merge_hyperparameters(hyperparameters, estimator.hyperparameters())
             parameters['HyperParameters'] = hyperparameters
 
         if experiment_config is not None:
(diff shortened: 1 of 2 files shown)


src/stepfunctions/steps/sagemaker.py
```diff
@@ -173,12 +173,7 @@ def __merge_hyperparameters(self, training_step_hyperparameters, estimator_hyper
             estimator_hyperparameters (dict): Hyperparameters specified in the estimator
         """
         merged_hyperparameters = estimator_hyperparameters.copy()
-        for key, value in training_step_hyperparameters.items():
-            if key in merged_hyperparameters:
-                logger.info(
-                    f"hyperparameter property: <{key}> with value: <{merged_hyperparameters[key]}> provided in the"
-                    f" estimator will be overwritten with value provided in constructor: <{value}>")
-            merged_hyperparameters[key] = value
+        merge_dicts(merged_hyperparameters, training_step_hyperparameters)
```
wong-a, line 176: Doesn't this do the same thing as before? The only difference seems to be `merge_dicts` does a deep copy, but that shouldn't be an issue. Seems like merging was already supported before. HyperParameters doesn't have nested fields.

I don't think this actually addresses https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/152, where the requester wanted to pass a Placeholder to `hyperparameters` arg, not a dict containing Placeholders as values.

ca-nguyen: Merging was only supported for non Placeholder values before. We would only merge if the constructor hyperparameters was a dict as well, otherwise, we would overwrite with the Estimator hyperparameters

ca-nguyen: I had also considered going back to what was done before the merge fix - if the hyperparameter passed in the constructor was a Placeholder, it would overwrite the estimator hyperparameters. But this would not satisfy the need for merging constructor and estimator hyperparameters.

Any suggestions?

wong-a: > I had also considered going back to what was done before the merge fix - if the hyperparameter passed in the constructor was a Placeholder, it would overwrite the estimator hyperparameters. But this would not satisfy the need for merging constructor and estimator hyperparameters.

I think we need to dive a little deeper here to understand the use case and what works today.

There are 3 scenarios:
1. `hyperparameters` is `None` - we use the hyperparameters from the estimator (already handled in v2.2.0)
2. `hyperparameters` is a dict (which may have Placeholder values) - we merge constructor's `hyperparameters` with the estimator's hyper parameters (already handled in v2.2.0)
3. `hyperparameters` is a Placeholder, not a dict with Placeholder values (raises Error)

[In the initial comment](https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/152#issue-943304296), the requester is trying to use the following, which is case 3:
```python
hyperparameters=execution_input["TrainingParameters"],
```

In 2.2.0 and this version, it doesn't work, because ExecutionInput doesn't have an `items()` method because it isn't a `dict`. This raises an Error:
```
AttributeError: 'ExecutionInput' object has no attribute 'items'
```

They used the following to convert it to a dict, but that doesn't work either.
```python
hyperparameters=execution_input["TrainingParameters"].get_schema_as_dict(),
```

The result of execution_input["TrainingParameters"].get_schema_as_dict() is the schema, not the placeholder values. e.g.
```python
{
   'C': str,
   'D': str
}
```

I think @ohbuchim is actually trying to merge `$$.Execution.Input.TrainingParameters` with the estimator's hyperparameters. There are two things we can do there.

1. @shivlaks' [proposal to just use the JSONPath for HypeterParameters (totally ignoring the estimator)](https://github.com/aws/aws-step-functions-data-science-sdk-python/issues/152#issuecomment-879338006)

```
'HyperParameters.$': "$$.Execution.Input['TrainingParameters']"
```

2. Try to merge the two by deriving values from the PlaceHolder schema. This requires all keys to be modelled in the Placeholder schema.

Suppose the estimator has these hyperparameters:
```python
{
   'A': 'a',
   'B': 'b'
}
```

And the ExecutionInput schema is:
```python
{
  'TrainingInput': {
     'B': str,
     'C': str
   },
}
```

If the caller passes `hyperparameters={ 'B': execution_input['TrainingInput']['B'], 'C': execution_input['TrainingInput']['C']` to `TrainingStep`, then we can already merge the two to become:
```python
{
   'A': 'a',
   'B.$': '$$.Execution.Input.B',
   'C.$': '$$.Execution.Input.C' 
}
```

If the caller passes `hyperparameters=execution_input['TrainingInput']`, then we need to handle this differently. `B` and `C` are not specified, so we would have to get all properties of the Placeholder schema at execution_input['TrainingInput'] and insert those keys into Parameters.

This requires the user to map all the properties they want to use in the schema. In ASL it's impossible to merge objects in cases where TrainingInput might have some arbitrary properties at runtime (e.g. D and E) not modelled in the schema.

ca-nguyen: >How so? The previous code doesn't do any type checks.
When a Placeholder was passed, an exception would be thrown due to this [line](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L176) since Placeholder does not have `items()` attribute (rectifying my comment: it would not replace it with the estimator hyperparameters, but would through an error)

ca-nguyen: (I just read the following comment - please ignore previous one)^

wong-a: With this change, scenario 3 will not throw an error, because of the type check in `merge_dicts`:
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/01e18c3777554882d1efc1410ecdcf25c4ec44e8/src/stepfunctions/steps/utils.py#L60

But that means we will ignore the `hyperparameters` arg entirely.

ca-nguyen: I like the solution you are proposing - this will provide the expected behaviour described in #152 
Will make the changes in the next commit

ca-nguyen: This will be resolved in 2 steps:
1. _Fix the breaking change_:
   - We need to support the same behaviour as in V2.1.0 -> if a Placeholder is provided as input to `hyperparameters` in TrainingStep constructor, the estimator hyperparameters are overwritten

2. _Feature request (described in #152 comments): To be able to merge a Placeholder hyperparameters passed to TrainingStep with the estimator hyperparameters. 
This will introduce a breaking change since we need to support the behaviour described in V2.1.0. Proposing to add a utility that translates a Placeholder to a usable jsonpath dict (with $). That translated placeholder_dict can then be used as hyperparameters input to TrainingStep. With that, merging the constructor and estimator hyperparameters will be possible.
   - ex: A Placeholder with schema
```
{
  'TrainingInput': {
     'B': str,
     'C': str
   }
}
```
will output the following:
```
{
   'B.$': '$$.Execution.Input.B',
   'C.$': '$$.Execution.Input.C' 
}
```

tests/unit/test_sagemaker_steps.py
```diff
@@ -292,6 +293,84 @@ def test_training_step_creation_with_placeholders(pca_estimator):
             'TrialComponentDisplayName': 'Training'
         },
         tags=DEFAULT_TAGS,
+        hyperparameters=execution_input['HyperParameters']
```
shivlaks, line 296: clarifying: can any of these other properties be expressed as placeholders? - if so, let's plug the gap to prevent future regressions. Ideally we have a test that covers all placeholder properties and not just some. 

That was the issue that made this bug easy to introduce, so lets it harder to introduce a bug by strengthening tests where possible.

tests/unit/test_sagemaker_steps.py
```diff
@@ -313,11 +392,13 @@ def test_training_step_creation_with_placeholders(pca_estimator):
...
-                'feature_dim': '50000',
-                'num_components': '10',
-                'subtract_mean': 'True',
+                'HyperParamA.$': "$$.Execution.Input['HyperParamA']",
+                'HyperParamB.$': "$$.Execution.Input['HyperParamB']",
                 'algorithm_mode': 'randomized',
-                'mini_batch_size': '200'
+                'feature_dim': 50000,
+                'mini_batch_size': 200,
+                'num_components.$': "$$.Execution.Input['num_components']",
```
shivlaks, line 400: nit: almost didn't see that this was being tested. perhaps make it easier for future contributors by adding a comment either here or where we set the `num_components` - since the estimator code isn't close to the test (locality) and this file is gigantic, might be a useful clue for others who only look at tests.

shivlaks (approved):
🎉 I think this issue is a `fix` and not a `feat`. had a couple other small comments.
please also make sure that @wong-a has a look before merging.

shivlaks (dismissed):

tests/unit/test_sagemaker_steps.py
ca-nguyen: good idea! will do

tests/unit/test_sagemaker_steps.py
```diff
@@ -292,6 +293,84 @@ def test_training_step_creation_with_placeholders(pca_estimator):
...
+        data=execution_input['Data'],
+        output_data_config_path=execution_input['OutputPath'],
+        experiment_config={
+            'ExperimentName': 'pca_experiment',
+            'TrialName': 'pca_trial',
+            'TrialComponentDisplayName': 'Training'
+        },
+        tags=DEFAULT_TAGS,
+        hyperparameters={
+            'num_components': execution_input['num_components'],
```
ca-nguyen, line 370: This will override the value that was defined in the estimator hyperparameters (see estimator hyperparameters definition [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/tests/unit/test_sagemaker_steps.py#L56-L61))

tests/unit/test_sagemaker_steps.py
ca-nguyen: All properties that were documented as being compatible with Placeholders are included in this test

We have an open feature request (#117) to support Placeholders for all parameters for Sagemaker steps - these will be added to the test as they are made Placeholder compatible

tests/unit/test_sagemaker_steps.py
```diff
@@ -275,6 +275,7 @@ def test_training_step_creation_with_placeholders(pca_estimator):
     execution_input = ExecutionInput(schema={
         'Data': str,
         'OutputPath': str,
+        'HyperParameters': str
```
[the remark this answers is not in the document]
wong-a: Just to be clear, it's not this line that would catch the regression, it's line 296 that assigns `hyperparameters=execution_input['HyperParameters']`. It could by any JSONPath.

tests/unit/test_sagemaker_steps.py
wong-a: If you think this is important to call out now, please add it as a **code comment** for future readers 🙂

wong-a: You can make it clearer what you're testing here by scoping down the assertion to just HyperParameters. It's easy to overlook when the dict being compared is so big. All that other stuff is already tested.

```python
assert step.to_dict().Parameters.HyperParameters == {
  'HyperParamA.$': "$$.Execution.Input['HyperParamA']",
  'HyperParamB.$': "$$.Execution.Input['HyperParamB']",
  ...
}

```

src/stepfunctions/steps/sagemaker.py
```diff
@@ -69,9 +69,10 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
                 * (list[sagemaker.amazon.amazon_estimator.RecordSet]) - A list of
                     :class:`sagemaker.amazon.amazon_estimator.RecordSet` objects,
                     where each instance is a different channel of training data.
-            hyperparameters (dict, optional): Parameters used for training.
-                    Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
+            hyperparameters: Parameters used for training.
+                * (dict[str, str], optional) - Hyperparameters supplied will be merged with the Hyperparameters specified in the estimator.
```
wong-a, line 73: Keep as it was before. `dict[str, str]` is not true, since keys can be `Placeholders` or strings.

tests/unit/test_sagemaker_steps.py
shivlaks: i meant as a comment in the code, not the PR 😅

src/stepfunctions/steps/sagemaker.py
```diff
@@ -69,9 +69,10 @@ def __init__(self, state_id, estimator, job_name, data=None, hyperparameters=Non
...
                     If there are duplicate entries, the value provided through this property will be used. (Default: Hyperparameters specified in the estimator.)
+                * (Placeholder, optional) - Hyperparameters supplied will overwrite the Hyperparameters specified in the estimator.
```
wong-a, line 75:
```suggestion
                * (Placeholder, optional) - The TrainingStep will use the hyperparameters specified by the Placeholder's value instead of the hyperparameters specified in the estimator.
```

wong-a (changes requested):
Looks good, just some minor comments for clarity

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Agreed! will change it back in the next commit

tests/unit/test_sagemaker_steps.py
ca-nguyen: It will be in both in next commit :D

ca-nguyen: >You can make it clearer what you're testing here by scoping down the assertion to just HyperParameters. It's easy to overlook when the dict being compared is so big. All that other stuff is already tested.

Good idea! changes will be added in next commit!

tests/unit/test_sagemaker_steps.py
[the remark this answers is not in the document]
shivlaks: yep, tagged my comment to the wrong line but that is what i meant. exercising hyperparameters as a placeholder would have caught this regression

ca-nguyen:
>🎉 I think this issue is a fix and not a feat. had a couple other small comments.
Done!

>Looks good, just some minor comments for clarity 
Comments addressed :)

wong-a (approved):

shivlaks (approved):

PR #160, opened 2021-09-09
fix: environment variables specified in Model or Estimator are not passed through to SageMaker ModelStep

ca-nguyen:
*Issue #, if available:* #82

*Description of changes:*
**1. fix: environment variables are overwritten and not passed through to SageMaker ModelStep**
- The environment variables are not passed from the model to ModelStep parameters (they are overwritten [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/sagemaker.py#L296)) for models without instance type (models other than [sagemake.model.FrameworkModel](https://sagemaker.readthedocs.io/en/stable/api/inference/model.html?highlight=model#sagemaker.model.FrameworkModel))
- With this change, the Model env parameters are passed to the ModelStep.

*New bug uncovered*: 
**2. fix: env variables defined in the Estimator are not translated to Model when calling `TrainingStep.get_expected_model()`**
- Fixed by passing the env from the estimator to the expected Model
- Added a new estimator (`pca_estimator_with_env`) with defined env parameters to use in test and confirm that the env variables are passed through to the Model
- With this change, the env variables defined in the estimator are passed to the expected Model when calling `TrainingStep.get_expected_model()`

### DIFF
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -286,10 +286,20 @@ class ModelStep(Task):
             instance_type (str, optional): The EC2 instance type to deploy this Model to. For example, 'ml.p2.xlarge'.
             tags (list[dict], optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
         """
-        if isinstance(model, Model):
+        if isinstance(model, FrameworkModel):
             parameters = model_config(model=model, instance_type=instance_type, role=model.role, image_uri=model.image_uri)
             if model_name:
                 parameters['ModelName'] = model_name
+        elif isinstance(model, Model):
+            parameters = {
+                'ExecutionRoleArn': model.role,
+                'ModelName': model_name or model.name,
+                'PrimaryContainer': {
+                    'Environment': model.env,
+                    'Image': model.image_uri,
+                    'ModelDataUrl': model.model_data
+                }
+            }
         else:
             raise ValueError("Expected 'model' parameter to be of type 'sagemaker.model.Model', but received type '{}'".format(type(model).__name__))
 
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -157,6 +157,8 @@ class TrainingStep(Task):
             model.name = model_name
         else:
             model.name = self.job_name
+        if self.estimator.environment is not None:
+            model.env = self.estimator.environment
         model.model_data = self.output()["ModelArtifacts"]["S3ModelArtifacts"]
         return model
 
@@ -284,20 +286,10 @@ class ModelStep(Task):
             instance_type (str, optional): The EC2 instance type to deploy this Model to. For example, 'ml.p2.xlarge'.
             tags (list[dict], optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
         """
-        if isinstance(model, FrameworkModel):
+        if isinstance(model, Model):
             parameters = model_config(model=model, instance_type=instance_type, role=model.role, image_uri=model.image_uri)
             if model_name:
                 parameters['ModelName'] = model_name
-        elif isinstance(model, Model):
-            parameters = {
-                'ExecutionRoleArn': model.role,
-                'ModelName': model_name or model.name,
-                'PrimaryContainer': {
-                    'Environment': {},
-                    'Image': model.image_uri,
-                    'ModelDataUrl': model.model_data
-                }
-            }
         else:
             raise ValueError("Expected 'model' parameter to be of type 'sagemaker.model.Model', but received type '{}'".format(type(model).__name__))
 
--- a/src/stepfunctions/steps/sagemaker.py
+++ b/src/stepfunctions/steps/sagemaker.py
@@ -157,7 +157,7 @@ class TrainingStep(Task):
             model.name = model_name
         else:
             model.name = self.job_name
-        if self.estimator.environment is not None:
+        if self.estimator.environment:
             model.env = self.estimator.environment
         model.model_data = self.output()["ModelArtifacts"]["S3ModelArtifacts"]
         return model
--- a/tests/unit/test_sagemaker_steps.py
+++ b/tests/unit/test_sagemaker_steps.py
@@ -50,6 +50,33 @@ def pca_estimator():
         role=EXECUTION_ROLE,
         instance_count=1,
         instance_type='ml.c4.xlarge',
+        output_path=s3_output_location
+    )
+
+    pca.set_hyperparameters(
+        feature_dim=50000,
+        num_components=10,
+        subtract_mean=True,
+        algorithm_mode='randomized',
+        mini_batch_size=200
+    )
+
+    pca.sagemaker_session = MagicMock()
+    pca.sagemaker_session.boto_region_name = 'us-east-1'
+    pca.sagemaker_session._default_bucket = 'sagemaker'
+
+    return pca
+
+
+@pytest.fixture
+def pca_estimator_with_env():
+    s3_output_location = 's3://sagemaker/models'
+
+    pca = sagemaker.estimator.Estimator(
+        PCA_IMAGE,
+        role=EXECUTION_ROLE,
+        instance_count=1,
+        instance_type='ml.c4.xlarge',
         output_path=s3_output_location,
         environment={
             'JobName': "job_name",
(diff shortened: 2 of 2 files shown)


tests/unit/test_sagemaker_steps.py
```diff
@@ -489,7 +493,10 @@ def test_training_step_creation_with_model(pca_estimator):
             'ExecutionRoleArn': EXECUTION_ROLE,
             'ModelName.$': "$['TrainingJobName']",
             'PrimaryContainer': {
-                'Environment': {},
+                'Environment': {
+                    'JobName': 'job_name',
+                    'ModelName': 'model_name'
+                },
```
shivlaks, line 499: question: typically, i'm not a huge fan of modifying existing tests as it can mask regressions. In this case it seems reasonable, but I think we should also have a test where environment is not established in the estimator to verify that it still renders to empty. 

you're currently testing the `if` part of the control flow when environment is set, but are also taking away the test of a scenario where it isn't.

wdyt?

shivlaks (changes requested):
thanks for putting this fix together

tests/unit/test_sagemaker_steps.py
ca-nguyen: There are tests that use a model without set environment variables (for example: [test_model_step_creation()](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/tests/unit/test_sagemaker_steps.py#L804) is using [pca_model](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/tests/unit/test_sagemaker_steps.py#L149) which doesn't have set environment variables)

I do agree that adding a test that tests the specific change would be better - will add it in next commit

shivlaks (approved):

yuan-bwn (dismissed):

src/stepfunctions/steps/sagemaker.py
```diff
@@ -284,20 +286,10 @@ def __init__(self, state_id, model, model_name=None, instance_type=None, tags=No
             instance_type (str, optional): The EC2 instance type to deploy this Model to. For example, 'ml.p2.xlarge'.
             tags (list[dict], optional): `List to tags <https://docs.aws.amazon.com/sagemaker/latest/dg/API_Tag.html>`_ to associate with the resource.
         """
-        if isinstance(model, FrameworkModel):
+        if isinstance(model, Model):
             parameters = model_config(model=model, instance_type=instance_type, role=model.role, image_uri=model.image_uri)
             if model_name:
                 parameters['ModelName'] = model_name
-        elif isinstance(model, Model):
```
wong-a, line 291: Is this backwards compatible?

ca-nguyen: yes, this generated the same parameters as the old way of doing, but takes into account the env variables as well

wong-a: Not quite. If `model` is the `Model` base class, parameters will have new fields that weren't there before. It's not just `PrimaryContainer.Evnironment`, but `VpcConfig` and other fields in the [container definition](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ContainerDefinition.html).

https://github.com/aws/sagemaker-python-sdk/blob/e78d0ead32ebd163a21996be44c8cd9a5ee31379/src/sagemaker/workflow/airflow.py#L566-L605

I don't have enough context to know what effect that has, but instantiating a ModelStep with the same model will not produce the same parameters as before.

ca-nguyen: You are right - it shouldn't produce the same parameters if `vpc_config` or an `image_config` (see [here](https://github.com/aws/sagemaker-python-sdk/blob/e78d0ead32ebd163a21996be44c8cd9a5ee31379/src/sagemaker/session.py#L4195-L4224)) were provided in the Model as they would be included in the parameters as well.

ca-nguyen: It is a good point to consider why those params were initially omitted, but I think it makes sense to have consistency between parameters generated from FrameworkModel with the ones generated from Model

With a FrameworkModel, params would also include `vpc_config`, but not `image_config` (container_def() is called without `image_config` arg [here](https://github.com/aws/sagemaker-python-sdk/blob/e78d0ead32ebd163a21996be44c8cd9a5ee31379/src/sagemaker/workflow/airflow.py#L563))

shivlaks: > it shouldn't produce the same parameters if vpc_config or an image_config (see here) were provided 

can we add the tests that *would* have caught this. our coverage might be limited 

> It is a good point to consider why those params were initially omitted,

Do the commits where they were introduced provide that context? In any case, we need to preserve existing behaviour. Customers who upgrade without changing any of their code should be able to do that without unexpected mutations.

src/stepfunctions/steps/sagemaker.py
```diff
@@ -284,20 +286,10 @@ def __init__(self, state_id, model, model_name=None, instance_type=None, tags=No
...
-            parameters = {
-                'ExecutionRoleArn': model.role,
-                'ModelName': model_name or model.name,
-                'PrimaryContainer': {
-                    'Environment': {},
```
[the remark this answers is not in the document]
ca-nguyen: I opted for this change since I saw it was supported in sagemaker sdk since the issue was opened, but I agree that we should not introduce breaking changes - even more if the issue does not capture the need to.

I'll revert the changes and go for the solution that was proposed in #84 and add test to validate that behaviour is the same

src/stepfunctions/steps/sagemaker.py
ca-nguyen: >can we add the tests that would have caught this. our coverage might be limited

Yes, i will add test to ensure there is no regression and/or breaking changes

>Do the commits where they were introduced provide that context? In any case, we need to preserve existing behaviour. Customers who upgrade without changing any of their code should be able to do that without unexpected mutations.

It was included in the initial commit, so no insight on why they were omitted

src/stepfunctions/steps/sagemaker.py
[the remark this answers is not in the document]
ca-nguyen: >edit: do we have tests that pass in a FrameworkModel and a Model?

We have tests that use a FrameworkModel (ex: [test_training_step_creation_with_framework()](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/tests/unit/test_sagemaker_steps.py#L503) and others that use a Model (ex: [test_training_step_creation_with_model()](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/tests/unit/test_sagemaker_steps.py#L449)), but none that pass both.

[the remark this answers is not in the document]
shivlaks: > but none that pass both.

I didn't mean in a single test (is that even possible). was just verifying that both parts of the control flow are being tested. I would assume that if they already existed, one of them would have broken with the attempt to introduce a breaking change.

if that didn't happen, i think it surfaces a gap in testing and we should use this opportunity to plug it in

tests/unit/test_sagemaker_steps.py
```diff
@@ -156,6 +191,31 @@ def pca_model():
     )
 
 
+@pytest.fixture
+def pca_model_with_env():
```
ca-nguyen, line 195: I've added a test using this Model that has defined env, vpc_config and imagie_config

tests/unit/test_sagemaker_steps.py
```diff
@@ -537,6 +597,63 @@ def test_training_step_creation_with_model(pca_estimator):
     }
 
 
+@patch('botocore.client.BaseClient._make_api_call', new=mock_boto_api_call)
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_training_step_creation_with_model_with_env(pca_estimator_with_env):
+    training_step = TrainingStep('Training', estimator=pca_estimator_with_env, job_name='TrainingJob')
```
ca-nguyen, line 603: This test is using a new model with a defined env, vpc_config and image_config 
vpc_config and image_config are not passed to ModelStep - this confirms that it is consistent with the current behaviour and will allow us to catch a breaking change.

src/stepfunctions/steps/sagemaker.py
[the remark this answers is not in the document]
ca-nguyen: I added a test in the latest commit that confirms we are consistent
The test will help catch future breaking changes

src/stepfunctions/steps/sagemaker.py
```diff
@@ -159,6 +159,8 @@ def get_expected_model(self, model_name=None):
             model.name = model_name
         else:
             model.name = self.job_name
+        if self.estimator.environment:
+            model.env = self.estimator.environment
```
shivlaks, line 163: CR description should also capture why we're doing this. It's not strictly related to the bug reported, but is solving an issue. Typically better to fix things separately, but when there are multiple fixes, they need to be captured in the commit summary, along with rationale and any testing that was performed.

ca-nguyen: Updated the description with the uncovered bug and what was done to fix it
I am open to move this fix to a separate PR - will go with what reviewers prefer since the review process has already begun

What do you think? @shivlaks  @wong-a

shivlaks: Given where we currently are, i'm not strongly opinionated because it feels more practical to go with it. It feels like it might just be busy work splitting them up in the current state.

Generally and going forward, I think we should keep changes contained and specific to the bug they address / feature they introduce for a few reasons:
* reverts are simpler if there's a problem with one fix and not with another
* the discovered bug may not have been triaged with a process of repro steps, expected behaviour, actual behaviour to converge on a candidate fix.
* simpler for reviewers and contributors - simplifies life for everyone involved
* helps to clearly see what was fixed for inclusion in changelog when we perform next release.
* this PR grows the scope of the change that it supersedes

tests/unit/test_sagemaker_steps.py
wong-a: _(**not blocking:** but please follow this guidance for future contributions)_

You're following the existing conventions here, but let's try to structure the tests as documentation so they are easier to read in the future.
1. Name the test cases in a way that describes the functionality you are verifying. e.g. `test_training_step_get_expected_model_returns_model_with_environment`
2. Scope down the assertions to specific pieces if you're testing specific things. e.g. `assert training_step.get_expected_model(...).env == # the environment`. You can do have multiple and smaller asserts in a single test. 
3. If something is already covered in other test cases, you don't to repeat it every time.

tests/unit/test_sagemaker_steps.py
wong-a: If this is important, write it in a code comment

src/stepfunctions/steps/sagemaker.py
wong-a: If it's already implemented and works, let's keep it in here, since it's a small enough change. Just update the PR description accordingly.

wong-a (approved):

tests/unit/test_sagemaker_steps.py
ca-nguyen: ACK
I'll take note to add these conventions in the CONTRIBUTING guide as well to ensure contributors have access to these

src/stepfunctions/steps/sagemaker.py
ca-nguyen: Done :)

shivlaks (approved):

PR #162, opened 2021-09-10
chore: add retry to SageMaker steps in integration tests

shivlaks:
#### Summary

Currently, we observe a few different failures that occur during integration tests, which get
executed as a part of the PR build as well as pushes to branches.

`createModel` and `createEndpoint` particularly see failures most frequently and they are
primarily:
* Rate exceeded  - `ThrottlingException`.

This change defines a default retry strategy that makes 5 attempts, over an interval of 5
seconds, which backs off with a multiplier of 2. The methodology behind this strategy is
naive and  may need some calibration. It should reduce the frequency of failures in the
short term.

We can adjust  the retry strategy as we go and expand to something more API specific as
the need arises.

#### Testing

* ran integ tests a few times locally - ensured they had the retry in the ASL definition and
executed through successfully.

rendered retry from the StateMachine definition on sagemaker steps:
```
"Retry": [
        {
          "ErrorEquals": [
            "SageMaker.AmazonSageMakerException"
          ],
          "IntervalSeconds": 5,
          "MaxAttempts": 5,
          "BackoffRate": 2
        }
      ]
```

* also pushed some dummy / trivial commits to this PR to trigger simultaneous builds. haven't
seen any state machine failures yet 🤞
---

### DIFF
--- a/tests/integ/__init__.py
+++ b/tests/integ/__init__.py
@@ -14,5 +14,15 @@ from __future__ import absolute_import
 
 import os
 
+from stepfunctions.steps import Retry
+
 DATA_DIR = os.path.join(os.path.dirname(__file__), "..", "data")
 DEFAULT_TIMEOUT_MINUTES = 25
+
+# Constants retrying on sagemaker exceptions when running tests
+SAGEMAKER_RETRY_STRATEGY = Retry(
+    error_equals=["SageMaker.AmazonSageMakerException"],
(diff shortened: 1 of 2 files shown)


ca-nguyen (approved):
Awesome work! 
This will allow us to push changes without worrying about triggering a build failures!

wong-a (approved):

Issue #125, opened 2021-03-19
Feature request: nested Step Functions


kmiller96:
Hi! I love your package. Until now I've been programmatically generating Step Function definitions using my own library (which was pretty hacky), so I love being able to migrate over to a better written one instead.

I think a really handy feature would be including nested Step Functions as a compute task. This is a common pattern that I use to help manage complexity and reuse high-level execution flows. This functionality is fully supported in AWS Step Functions already: https://docs.aws.amazon.com/step-functions/latest/dg/connect-stepfunctions.html

If you can point me in the right direction I'm happy to make the contribution to the source myself.

wong-a:
Glad you're liking it! All state types and service integration steps are defined in [src/stepfunctions/steps/](https://github.com/aws/aws-step-functions-data-science-sdk-python/tree/main/src/stepfunctions/steps). 

Here's an example of a Lambda step:
https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/compute.py#L19-L45

It's fairly trivial to add new service integrations. We might consider some additional functionality to make it easier to use with `Workflow` instances, but all you'd really need to pass is the state machine ARN and execution input.

tomaszdudek7:
Hello @shivlaks. I see that this issue is in progress. Are there any chances of merging it any time soon?

shivlaks:
hey @tomaszdudek7 - intending to have a PR out for this this week and targeting the next released version to include this support!

tomaszdudek7:
Great to hear.

shivlaks:
@tomaszdudek7 - our wonderful team member @ca-nguyen will be implementing, but targets remain the same - shooting for end-of-week PR and shipping with the next release

AustinGomez:
Hi, any updates here? I saw that the PR has been opened but haven't seen anything since. Thank you!

ca-nguyen:
Hi @AustinGomez,
We are aiming to have traction on the PR and merge the fix by end of week next week
Thanks for your interest in this feature!

PR #156, opened 2021-08-17
feat: Add support for Amazon EKS

ca-nguyen:
*Issue #, if available:* #106

*Description of changes:*
Added:
- [EksRunJob](https://docs.aws.amazon.com/step-functions/latest/dg/connect-eks.html) to create a Task state that allows you to run a job on your Amazon EKS cluster
- [EksCall](https://docs.aws.amazon.com/step-functions/latest/dg/connect-eks.html) to create a Task state that allows you to use the Kubernetes API to read and write Kubernetes resource objects via a Kubernetes API endpoint
- [EksCreateClusterStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html) to create a Task state that creates an Amazon EKS cluster
- [EksDeleteClusterStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_DeleteCluster.html) to create a Task state that deletes an Amazon EKS cluster
- [EksCreateFargateProfileStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateFargateProfile.html) to create a Task state that creates a Fargate profile
- [EksDeleteFargateProfileStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_DeleteFargateProfile.html) to create a Task state that deletes a Fargate profile
- [EksCreateNodeGroupStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_DeleteCluster.html) to create a Task state that creates a node group
 - [EksDeleteNodeGroupStep](https://docs.aws.amazon.com/eks/latest/APIReference/API_DeleteNodegroup.html) to create a Task state that deletes a node group

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #151, opened 2021-07-08
feat: Add support for AWS Glue DataBrew

ca-nguyen:
*Issue #, if available:* #112

*Description of changes:*
Added a [StartJobRun](https://docs.aws.amazon.com/databrew/latest/dg/API_StartJobRun.html) step which allows to run a DataBrew job. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #166, opened 2021-09-11
Add support for nested Aws StepFunctions service integration

ca-nguyen:
*Issue #, if available:* #125

*Description of changes:*
Added support for [StartExecution](https://docs.aws.amazon.com/step-functions/latest/dg/connect-stepfunctions.html) step that allows to start an execution of another state machine

*Testing*
- Added unit tests
- Manual tests for all resources:
   - "arn:aws:states:::states:startExecution"
   - "arn:aws:states:::states:startExecution.sync:2"
   - "arn:aws:states:::states:startExecution.sync"
   - "arn:aws:states:::states:startExecution.waitForTaskToken"

### DIFF
--- a/src/stepfunctions/steps/integration_resources.py
+++ b/src/stepfunctions/steps/integration_resources.py
@@ -25,6 +25,7 @@ class IntegrationPattern(Enum):
     WaitForTaskToken = "waitForTaskToken"
     WaitForCompletion = "sync"
     RequestResponse = ""
+    WaitForCompletionWithJsonResponse = "sync:2"
 
 
 def get_service_integration_arn(service, api, integration_pattern=IntegrationPattern.RequestResponse):
--- a/src/stepfunctions/steps/service.py
+++ b/src/stepfunctions/steps/service.py
@@ -24,6 +24,7 @@ EVENTBRIDGE_SERVICE_NAME = "events"
 GLUE_DATABREW_SERVICE_NAME = "databrew"
 SNS_SERVICE_NAME = "sns"
 SQS_SERVICE_NAME = "sqs"
+STEP_FUNCTIONS_SERVICE_NAME = "states"
 
 
 class DynamoDBApi(Enum):
@@ -68,6 +69,10 @@ class SnsApi(Enum):
 
 class SqsApi(Enum):
     SendMessage = "sendMessage"
+
+
+class StepFunctions(Enum):
+    StartExecution = "startExecution"
 
 
 class DynamoDBGetItemStep(Task):
@@ -887,3 +892,59 @@ class EmrModifyInstanceGroupByNameStep(Task):
                                                                    ElasticMapReduceApi.ModifyInstanceGroupByName)
 
         super(EmrModifyInstanceGroupByNameStep, self).__init__(state_id, **kwargs)
+
+
+class StepFunctionsStartExecutionStep(Task):
+
+    """
+    Creates a Task state that starts an execution of another state machine. See `Manage AWS Step Functions Executions as an Integrated Service <https://docs.aws.amazon.com/step-functions/latest/dg/connect-stepfunctions.html`_ for more details.
+    """
+
+    def __init__(self, state_id, wait_for_callback=False, wait_for_completion=True, json_ouput=False, **kwargs):
+        """
+        Args:
+            state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
+            wait_for_callback(bool, optional): Boolean value set to `True` if the Task state should wait for callback to resume the operation. (default: False)
+            timeout_seconds (int, optional): Positive integer specifying timeout for the state in seconds. If the state runs longer than the specified timeout, then the interpreter fails the state with a `States.Timeout` Error Name. (default: 60)
+            timeout_seconds_path (str, optional): Path specifying the state's timeout value in seconds from the state input. When resolved, the path must select a field whose value is a positive integer.
+            heartbeat_seconds (int, optional): Positive integer specifying heartbeat timeout for the state in seconds. This value should be lower than the one specified for `timeout_seconds`. If more time than the specified heartbeat elapses between heartbeats from the task, then the interpreter fails the state with a `States.Timeout` Error Name.
+            heartbeat_seconds_path (str, optional): Path specifying the state's heartbeat value in seconds from the state input. When resolved, the path must select a field whose value is a positive integer.
+            comment (str, optional): Human-readable comment or description. (default: None)
+            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            parameters (dict, optional): The value of this field becomes the effective input for the state.
+            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+            wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait to complete before proceeding to the next step in the workflow. (default: True)
+            json_output(bool, optional): Boolean value set to `True` if the the Task should return a response in JSON format. If set to `False`, the Task should return a str response (default: False)
+        """
+        if wait_for_callback:
+            """
+            Example resource arn: arn:aws:states:::states:startExecution.waitForTaskToken
+            """
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution,
+                                                                       IntegrationPattern.WaitForTaskToken)
+        elif wait_for_completion:
+            if json_ouput:
+                """
+                Example resource arn:aws:states:::states:startExecution.sync:2
+                """
+                kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution,
+                                                                       IntegrationPattern.WaitForCompletionWithJsonResponse)
+            else:
+                """
+                Example resource arn:aws:states:::states:startExecution.sync
+                """
+                kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution,
+                                                                       IntegrationPattern.WaitForCompletion)
+        else:
+            """
+            Example resource arn:aws:states:::states:startExecution
+            """
+            kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution)
+
+        super(StepFunctionsStartExecutionStep, self).__init__(state_id, **kwargs)
--- a/tests/unit/test_service_steps.py
+++ b/tests/unit/test_service_steps.py
@@ -30,6 +30,7 @@ from stepfunctions.steps.service import EmrCreateClusterStep, EmrTerminateCluste
 from stepfunctions.steps.service import EventBridgePutEventsStep
 from stepfunctions.steps.service import SnsPublishStep, SqsSendMessageStep
 from stepfunctions.steps.service import GlueDataBrewStartJobRunStep
+from stepfunctions.steps.service import StepFunctionsStartExecutionStep
 
 
 @patch.object(boto3.session.Session, 'region_name', 'us-east-1')
@@ -1157,4 +1158,109 @@ def test_eks_call_step_creation():
             }
         },
         'End': True
+    }
+
+
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_step_functions_start_execution_step_creation():
+    step = StepFunctionsStartExecutionStep(
+        "SFN Start Execution", wait_for_callback=False, wait_for_completion=False,
+        parameters={
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        })
+
+    assert step.to_dict() == {
+        "Type": "Task",
+        "Resource": "arn:aws:states:::states:startExecution",
+        "Parameters": {
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        },
+        "End": True
+    }
+
+
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_step_functions_start_execution_step_creation_sync_json_response():
+    step = StepFunctionsStartExecutionStep(
+        "SFN Start Execution - Sync with json response", wait_for_callback=False, wait_for_completion=True,
+        json_ouput=True, parameters={
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        })
+
+    assert step.to_dict() == {
+        "Type": "Task",
+        "Resource": "arn:aws:states:::states:startExecution.sync:2",
+        "Parameters": {
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        },
+        "End": True
+    }
+
+
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_step_functions_start_execution_step_creation_sync_json_response():
+    step = StepFunctionsStartExecutionStep(
+        "SFN Start Execution - Sync with string response", wait_for_callback=False, wait_for_completion=True,
+        json_ouput=False, parameters={
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        })
+
+    assert step.to_dict() == {
+        "Type": "Task",
+        "Resource": "arn:aws:states:::states:startExecution.sync",
+        "Parameters": {
+            "Input": {
+                "Comment": "Hello world!"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        },
+        "End": True
+    }
+
+
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_step_functions_start_execution_step_creation_wait_for_callback():
+    step = StepFunctionsStartExecutionStep(
+        "SFN Start Execution - Wait for Callback", wait_for_callback=True, parameters={
+            "Input": {
+                "Comment": "Hello world!",
+                "token.$": "$$.Task.Token"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        })
+
+    assert step.to_dict() == {
+        "Type": "Task",
+        "Resource": "arn:aws:states:::states:startExecution.waitForTaskToken",
+        "Parameters": {
+            "Input": {
+                "Comment": "Hello world!",
+                "token.$": "$$.Task.Token"
+            },
+            "StateMachineArn": "arn:aws:states:us-east-1:123456789012:stateMachine:HelloWorld",
+            "Name": "ExecutionName"
+        },
+        "End": True
     }
--- a/doc/services.rst
+++ b/doc/services.rst
@@ -20,6 +20,8 @@ This module provides classes to build steps that integrate with Amazon DynamoDB,
 
 - `Amazon SQS <#amazon-sqs>`__
 
+- `AWS Step Functions <#aws-step-functions>`__
+
 
 Amazon DynamoDB
 ----------------
@@ -82,3 +84,8 @@ Amazon SNS
 Amazon SQS
 -----------
 .. autoclass:: stepfunctions.steps.service.SqsSendMessageStep
+
+AWS Step Functions
+------------------
+.. autoclass:: stepfunctions.steps.service.StepFunctionsStartExecutionStep
+
--- a/src/stepfunctions/steps/__init__.py
+++ b/src/stepfunctions/steps/__init__.py
@@ -34,3 +34,4 @@ from stepfunctions.steps.service import EmrCreateClusterStep, EmrTerminateCluste
 from stepfunctions.steps.service import EventBridgePutEventsStep
 from stepfunctions.steps.service import GlueDataBrewStartJobRunStep
 from stepfunctions.steps.service import SnsPublishStep, SqsSendMessageStep
+from stepfunctions.steps.service import StepFunctionsStartExecutionStep


src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,59 @@ def __init__(self, state_id, **kwargs):
...
+                """
+                kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution,
+                                                                       IntegrationPattern.WaitForCompletion)
+        else:
+            """
+            Example resource arn:aws:states:::states:startExecution
+            """
+            kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution)
```
shivlaks, line 948: suggestion: should there be a flag to ask for this? 

Currently, Asking for this property is also counter-intuitive because it's a behaviour that is achieved by setting `wait_for_callback` and `wait_for_completion` to false. 

A customer saying "no" to these 2 properties != a "yes" for the behaviour they will get. Additionally, this behaviour is not currently documented in any way.

Recommendation:
* Add a flag to opt into the behaviour. Clearly document what happens when that flag is set to `True`
  *   Alternatively could model these 3 properties in a more complex type, but I'm not leaning that way because that might be disruptive and too different from usage of other service integrations.

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,59 @@ def __init__(self, state_id, **kwargs):
...
+
+    """
+    Creates a Task state that starts an execution of another state machine. See `Manage AWS Step Functions Executions as an Integrated Service <https://docs.aws.amazon.com/step-functions/latest/dg/connect-stepfunctions.html`_ for more details.
+    """
+
+    def __init__(self, state_id, wait_for_callback=False, wait_for_completion=True, json_ouput=False, **kwargs):
+        """
+        Args:
+            state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
+            wait_for_callback(bool, optional): Boolean value set to `True` if the Task state should wait for callback to resume the operation. (default: False)
```
shivlaks, line 907: nit: we should probably keep all the service integration pattern flags co-located in the docs. locality is valuable and easier to digest. I think we can do better by not having related properties (like wait_for_completion) documented 10 lines apart.

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,59 @@ def __init__(self, state_id, **kwargs):
...
+        """
+        if wait_for_callback:
+            """
+            Example resource arn: arn:aws:states:::states:startExecution.waitForTaskToken
+            """
+
+            kwargs[Field.Resource.value] = get_service_integration_arn(STEP_FUNCTIONS_SERVICE_NAME,
+                                                                       StepFunctions.StartExecution,
+                                                                       IntegrationPattern.WaitForTaskToken)
+        elif wait_for_completion:
```
shivlaks, line 928: missing validation:

if both `wait_for_callback` and `wait_for_completion` properties are set `True`, what the customer wants is unclear and we should raise an error. What will actually happen is we will *always* give them `wait_for_callback`

shivlaks (changes requested):
left some feedback of some changes I think we should make and some suggestions to consider. haven't looked at the tests yet.

src/stepfunctions/steps/service.py
ca-nguyen: I agree - will regroup them in the next commit

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,59 @@ def __init__(self, state_id, **kwargs):
...
+            comment (str, optional): Human-readable comment or description. (default: None)
+            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            parameters (dict, optional): The value of this field becomes the effective input for the state.
+            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+            wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait to complete before proceeding to the next step in the workflow. (default: True)
+            json_output(bool, optional): Boolean value set to `True` if the the Task should return a response in JSON format. If set to `False`, the Task should return a str response (default: False)
```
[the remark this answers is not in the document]
ca-nguyen: >- default to sync:2

Agreed - let's default to sync:2 - I think we should be consistent with the CDK and this seems like it would be the most used resource between the two.

>- don't make json_output configurable

Since `startExecution.sync` is in our documentation, I believe it could be good to provide an option for our customers to use it. How about instead of being `json_output` flag, we provide a flag to opt in for a string response instead? Something like `string_response` or `string_output` that can be set in the kwargs. I'm curious to see what other reviewers think too and will go with the majority

src/stepfunctions/steps/service.py
ca-nguyen: I like your suggestion - I agree, having to disable to properties to activate another is not intuitive. A flag might make things clearer!
I had thought of introducing an enum for the property type, but as you mentioned, the implementation would diverge from the other service integrations.
I can add a check to see if more than 1 is enabled and throw an exception if that is the case.

src/stepfunctions/steps/service.py
ca-nguyen: Agreed - will add that in the next commit

src/stepfunctions/steps/service.py
ca-nguyen: What would be the default if all flags are disabled? Should we return an error?
If we were to keep only the 2 initial property flags `wait_for_callback` and `wait_for_completion` and:
- Raise an exception if both were set to True (as you mentioned below) 
- Update documentation to illustrate how to select this behaviour (Disable both false to get arn:aws:states:::states:startExecution resource)

I think this way, would be clearer for the customers instead of having them enable at least 1 of the 3 flags - What do you think?

src/stepfunctions/steps/service.py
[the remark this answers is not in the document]
shivlaks: > Since startExecution.sync is in our documentation, I believe it could be good to provide an option for our customers to use it.

It will always be in the documentation as it was the only option available initially. As we refined our approach and offered a better alternative, `sync:2` was introduced. 

We shouldn't seek to hit arbitrarily high 1:1 parity with documentation. Rather, we should ask ourselves what use cases it unlocks, whether `sync:2` precludes the ability to do anything that `sync` offers. Do you have a use case in mind?

We have not exposed this integration to customers yet, so my recommendation is still to start as tight as possible and expand it in the future if we have a feature request / use case in mind. Nothing prevents us from exposing it in a future release, but supporting it at the outset based on assumptions is fragile and also a 1-way door as we'll have it out there forever (at least until v3).

thoughts?

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,74 @@ def __init__(self, state_id, **kwargs):
...
+            comment (str, optional): Human-readable comment or description. (default: None)
+            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            parameters (dict, optional): The value of this field becomes the effective input for the state.
+            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+        """
+        # Validate that only 1 one of the property flags are set to true
+        enabled_property_flags = []
+        property_flags = [(wait_for_callback, "wait_for_callback"), (wait_for_completion, "wait_for_completion")]
+        for property_flag in property_flags:
```
wong-a, line 926: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

To create a `list`, try to use `list` comprehension instead of a loop. List comprehension is the preferred way to make a list using Python, and it&#39;s simpler and easier to understand than using a loop.

[Learn more](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions)

src/stepfunctions/steps/service.py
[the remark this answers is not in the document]
ca-nguyen: >We have not exposed this integration to customers yet, so my recommendation is still to start as tight as possible and expand it in the future if we have a feature request / use case in mind. Nothing prevents us from exposing it in a future release, but supporting it at the outset based on assumptions is fragile and also a 1-way door as we'll have it out there forever (at least until v3).

I definitely agree with this - we shouldn't introduce 1 way doors. I don't have as specific use case in mind where `:sync` would be preferred over `:sync:2`. If there is ask for it, we can always add support for it later and it won't be a breaking change

src/stepfunctions/steps/service.py
shivlaks: > If we were to keep only the 2 initial property flags wait_for_callback and wait_for_completion

How would users use the async invocation method, this only enables 2/3 possible modes of invoking the service integration. I think we need to provide users with an intuitive way to ask for `arn:aws:states:::states:startExecution`.

I'm not in favour of setting `wait_for_callback` and `wait_for_completion` to false to achieve that. It's counter-intuitive to me as a developers, but perhaps others will disagree...


* the default if none of the properties are set should be to use `wait_for_completion` aka `sync:2`
  * Rationale: it's what all the other service integrations establish as a default
* if a user sets all the flags to False, then it's a validation error
  * Rationale: it's required that one of the modes of operation is established and the user's intention is unclear.

Aside:
These implementation decisions should be captured and elaborated in the PR summary itself. Right now the description only captures the outcome, but it's valuable context for posterity as well as helping reviewers understand why the code is structured the way it is. Please ensure you update the PR description with implementation decisions that are being made.

ca-nguyen: >- if a user sets all the flags to False, then it's a validation error
Rationale: it's required that one of the modes of operation is established and the user's intention is unclear.

Agreed - this will ensure the customer's intention is clear. 
I'll go with what you are proposing - we want to provide an intuitive solution for our customers

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,66 @@ def __init__(self, state_id, **kwargs):
...
+            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            parameters (dict, optional): The value of this field becomes the effective input for the state.
+            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+        """
+        # Validate that only 1 one of the property flags are set to true
+        property_flags = [(wait_for_callback, "wait_for_callback"), (wait_for_completion, "wait_for_completion"),
+                          (async_call, "async_call")]
+        enabled_property_flags = [property_flag[1] for property_flag in property_flags if property_flag[0]]
+        if len(enabled_property_flags) == 0:
```
wong-a, line 927: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

To check if a container or sequence (string, list, tuple) is empty, use `if not val`. Do not compare its length using `if len(val) == 0` or `if len(val) &gt; 0`

 [Learn more](https://www.python.org/dev/peps/pep-0008/#programming-recommendations#:~:text=if%20not%20seq)

src/stepfunctions/steps/service.py
ca-nguyen: >Please ensure you update the PR description with implementation decisions that are being made.

Updated the description

src/stepfunctions/steps/service.py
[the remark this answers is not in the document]
shivlaks: > we shouldn't introduce 1 way doors.

That's not quite what I meant. I meant that when we go through a 1-way door, we should have a deep understanding of **why** we are doing it.

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +892,66 @@ def __init__(self, state_id, **kwargs):
...
+        One of three must be enabled to create the step successfully.
+    """
+
+    def __init__(self, state_id, wait_for_callback=False, wait_for_completion=True, async_call=False, **kwargs):
+        """
+        Args:
+            state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
+            wait_for_callback(bool, optional): Boolean value set to `True` if the Task state should wait for callback to resume the operation. (default: False)
+            wait_for_completion (bool, optional): Boolean value set to `True` if the Task state should wait to complete before proceeding to the next step in the workflow. (default: True)
+            async_call(bool, optional): Boolean value set to `True` for the Task proceed to the next step in the workflow without waiting for completion. (default: False):
```
wong-a, line 912: This would be a good opportunity to improve as proposed here: https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/74

Using a single enum arg is easier to reason about instead of three mutually exclusive args, which either overwrite each other or raise an Error if more than one are provided.

shivlaks: I'm in favour of an enum as well (which is what we also do in the CDK) - the tradeoff is it might be dissimilar to other service integrations.

i.e. if any of the existing service integrations add support for all 3 modes of operation, we would likely have some inconsistency between service integrations.

not sure that it's a problem we need to futureproof for, but just mentioning it for completeness

ca-nguyen: Also in favour of an enum even though it introduces inconsistency with the other service integrations
This will be clearer and easier to use for our customers

wong-a: None of the service integrations today support all 3 patterns, which is why this is a good time to introduce this change.

`ecs:runTask` does support all 3, but this SDK only exposes `.sync` and request-response. 

- https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/compute.py#L184-L197
- https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html

shivlaks: this is the enum we added in the [CDK](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-stepfunctions.IntegrationPattern.html)
and Python specific [documentation](https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_stepfunctions/IntegrationPattern.html) for it

keep in mind that you'd want to name the options in the enum in a way that's idiomatic for existing data science SDK customers.

wong-a: We have an enum already https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/23878de4ba909564569b90d9ab3ae689cacf04dd/src/stepfunctions/steps/integration_resources.py#L20

ca-nguyen: I was planning to use the the existing enum mentioned by @wong-a 
Change to use enum arg instead of 3 flags will be included in the next commit

shivlaks: i should clarify:

wasn't suggesting adding a new enum, was only sharing the CDK enum to indicate that there shouldn't be a new entry in the enum to surface `sync:2` - it's just what `WaitForCompletion` maps to for this service integration.

> None of the service integrations today support all 3 patterns, which is why this is a good time to introduce this change.

That's true, but my comment around futureproofing is more around how you'd introduce it if a service integration that currently supports 2 patterns starts supporting all 3.

i.e. if EMR aded wait for completion, since we have already exposed `wait_for_completion` as a property:
* we could introduce the enum as input allowing for all 3 integration types
  * we would essentially map 2 properties as ways in which wait_for_completion could be requested (enum and flag)
  * we would need to add validation that enum and this flag don't conflict with each other

I still prefer the enum because it's clear and explicit about behaviour that will be observed.

ca-nguyen: >wasn't suggesting adding a new enum, was only sharing the CDK enum to indicate that there shouldn't be a new entry in the enum to surface sync:2 - it's just what WaitForCompletion maps to for this service integration.

Yes, i was planning to remove the `WaitForCompletionWithJsonResponse` that was added in this PR to not expose the response format to the customer (this was discussed [here](https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/166#discussion_r706701234))

shivlaks:
not directly related to this PR, but since we are trying to address the suggestion in #74 - any reason we shouldn't follow the same pattern for un-released service integrations we've added recently in #156 and #151 (not in this PR)

I feel we should establish that in all service integrations we support, but especially ones we have not released yet.

thoughts? - @wong-a  @ca-nguyen

ca-nguyen:
> not directly related to this PR, but since we are trying to address the suggestion in #74 - any reason we shouldn't follow the same pattern for un-released service integrations we've added recently in #156 and #151 (not in this PR)

Since these service integrations have not been released yet, it makes sense to follow the same patterns that provide a clearer solution to our customers. 

>I feel we should establish that in all service integrations we support, but especially ones we have not released yet.
I agree - ideally, we are consistent through all the service integrations

src/stepfunctions/steps/service.py
ca-nguyen: After reflexion, I opted to add a new enum so that we have the same terms than the ones described in [Service Integration Patterns](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html) similarly to the CDK doc mentioned by @shivlaks (details in PR description)

ca-nguyen: >i.e. if EMR aded wait for completion, since we have already exposed wait_for_task_token as a property:
> - we could introduce the enum as input allowing for all 3 integration types
>   - we would essentially map 2 properties as ways in which wait_for_task_token could be requested (enum and flag)
>   - we would need to add validation that enum and this flag don't conflict with each other
> I still prefer the enum because it's clear and explicit about behaviour that will be observed.
> If that situation materialized, I'd also recommend marking the flag as deprecated and dropping support in v3 to normalize every integration to using the enum.

This seems like an acceptable solution when adding support for a 3rd integration pattern
Mapping the existing flag to the enum and marking that flag as "to be deprecated" will ensure backward compatibility and give a heads up to our customers

For consistency, when adding support for a new service integrations, regardless of the number of integration patterns, I also think we should use the enum solution if we decide to go that route

src/stepfunctions/steps/integration_resources.py
```diff
@@ -25,6 +29,17 @@ class IntegrationPattern(Enum):
...
+    WaitForCompletionWithJsonResponse = "sync:2"
+
+
+class ServiceIntegrationType(Enum):
+    """
+    Service Integration Types for service integration resources (see `Service Integration Patterns <https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html`_ for more details.)
+    """
+
+    REQUEST_RESPONSE = "RequestResponse",
+    RUN_A_JOB = "RunAJob",
```
shivlaks, line 41: this is just called `RUN_JOB` in the CDK. I would think that we should try and stick to something similar?

shivlaks (changes requested):
not sure I follow this part tradeoffs:
>  An alternative mentioned by @shivlaks (in the discussion) is to use the existing flags to map to the ServiceIntegrationType and add a note in the docstrings that those flags will be deprecated in v3.0.0 and deprecate them when we are ready to upgrade the major version

I was not suggesting using existing flags to map to service integration type. I was indicating that we should *not* have default behaviour that a user does not explicitly opt into. The callout was that *if* a user achieves a certain behaviour, they need to explicitly opt into it.

I don't think this is a design decision at all and really the only tradeoff worth documenting is that we are opting into inconsistency for simplicity. The future consideration worth documenting is what we will do with existing service integrations as well as all future integrations. Will other service integrations now follow the enum pattern or fall back to past patterns?

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +893,60 @@ def __init__(self, state_id, **kwargs):
...
+            heartbeat_seconds_path (str, optional): Path specifying the state's heartbeat value in seconds from the state input. When resolved, the path must select a field whose value is a positive integer.
+            comment (str, optional): Human-readable comment or description. (default: None)
+            input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
+            parameters (dict, optional): The value of this field becomes the effective input for the state.
+            result_path (str, optional): Path specifying the raw input’s combination with or replacement by the state’s result. (default: '$')
+            output_path (str, optional): Path applied to the state’s output after the application of `result_path`, producing the effective output which serves as the raw input for the next state. (default: '$')
+        """
+        supported_integ_types = [ServiceIntegrationType.REQUEST_RESPONSE, ServiceIntegrationType.RUN_A_JOB,
```
shivlaks, line 923: you called the enum patterns

```suggestion
        supported_integration_patterns = [ServiceIntegrationType.REQUEST_RESPONSE, ServiceIntegrationType.RUN_A_JOB,
```

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +893,60 @@ def __init__(self, state_id, **kwargs):
...
+                                 ServiceIntegrationType.WAIT_FOR_CALLBACK]
+
+        if not isinstance(service_integration_type, ServiceIntegrationType):
+            raise ValueError(f"Invalid type used for service_integration_type arg ({service_integration_type}, "
+                             f"{type(service_integration_type)}). Accepted type: {ServiceIntegrationType}")
+        elif service_integration_type not in supported_integ_types:
+            raise ValueError(f"Service Integration Type ({service_integration_type.name}) is not supported for this step - "
+                             f"Please use one of the following: "
+                             f"{[integ_type.name for integ_type in supported_integ_types]}")
```
shivlaks, line 932: can we move the validation into it's own function and keep the constructor clean??
I imagine you'll want to reuse this validation across other service integrations as well.

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +893,60 @@ def __init__(self, state_id, **kwargs):
...
+
+        if service_integration_type == ServiceIntegrationType.RUN_A_JOB:
+            # For RunAJob service integration type, use the resource that returns json response (`.sync:2`)
```
shivlaks, line 935: nit: unnecessary comment - if we do want to document, we should capture the "why". the code already explains what it's doing.

tests/unit/test_service_steps.py
```diff
@@ -1158,3 +1161,92 @@ def test_eks_call_step_creation():
         },
         'End': True
     }
+
+
+@patch.object(boto3.session.Session, 'region_name', 'us-east-1')
+def test_step_functions_start_execution_step_creation():
```
shivlaks, line 1167: test name should include the integration pattern you're testing or "default" if you make it optional (as it is in the CDK)

src/stepfunctions/steps/service.py
```diff
@@ -887,3 +893,60 @@ def __init__(self, state_id, **kwargs):
...
+
+    Property flags: There are three flags (wait_for_callback, wait_for_completion and async_call) that can be set in order to select which Step Functions resource to use.
+        One of three must be enabled to create the step successfully.
+    """
+
+    def __init__(self, state_id, service_integration_type, **kwargs):
+        """
+        Args:
+            state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
+            integration_pattern (stepfunctions.steps.integration_resources.ServiceIntegrationType): Service integration type to use to build resource.
```
shivlaks, line 911: 1. should this be optional with a default?
2. this is not the name of your parameter

tests/unit/test_service_steps.py
```diff
@@ -1158,3 +1161,92 @@ def test_eks_call_step_creation():
...
+    step = StepFunctionsStartExecutionStep(
+        "SFN Start Execution", service_integration_type=ServiceIntegrationType.REQUEST_RESPONSE, parameters={
+            "Input": {
+                "Comment": "Hello world!"
```
shivlaks, line 1171: nit: is comment needed at all? - try to keep tests minimal. applies to all tests

shivlaks (changes requested):
getting closer and certainly looks a lot nicer with the integration pattern. 
take a look at my latest round of feedback and let me know if you have any questions

ca-nguyen:
>But no integ tests are in the diff. Did you forget to commit them? What kind of manual testing did you do?

I added details on the Manual tests that were performed in the `Testing` section of the PR summary.
No integ test were added since we needed an external Lambda function to test. This was checked since this was N/A and to reflect that no more tasks were left. The template says:

>Please check all boxes (including N/A items)

I updated with a Note next to the checked task to avoid confusion

PR #169, opened 2021-09-23
feat: Add retry and catch to Retriers and Catchers when passed to constructor

### LINKED ISSUE #115: Can't serialize graphs using Retry
Just installed v1.1.2

```python
from stepfunctions.steps import Graph, Task, states, Retry
retry = Retry(error_equals=["States.TaskFailed"])
Task(state_id="whatever", retry=retry).to_json()
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/fitz/Workspace/github.com/mobikitinc/data-ingest/venv/lib/python3.8/site-packages/stepfunctions/steps/states.py", line 93, in to_json
    return json.dumps(self.to_dict())
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Retry is not JSON serializable

```

ca-nguyen:
### Description

This will allow retry and catch blocks to be added directly from the constructor.

Fixes #115 

### Why is the change necessary?

Currently, `retry` and `catch` blocks passed to Task, Parallel and Map State constructors would not add those blocks to Retriers and Catchers. With this change, it will possible do so. 

Per the ASL documentation, Retry and Catch fields are arrays that contain Retriers and Catchers (see [Task ASL doc](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-task-state.html) for ex). This PR makes it possible to use lists or Retry/Catch objects in the constructor for `retry` and `catch` constructor arguments.

### Solution

In the constructor, when `retry`/`catch` is not `None`, add it to the state's retries/catches.
This change is done for the 3 States that support Retry and Catch blocks: [Task](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-task-state.html), [Parallel](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-parallel-state.html) and [Map](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html).

### Testing

- Added unit and integ tests
- Manual testing

----

### Pull Request Checklist

Please check all boxes (including N/A items)

#### Testing

- [X] Unit tests added
- [X] integration test added

#### Documentation

- [X] __docs__: All relevant [docs](https://github.com/aws/aws-step-functions-data-science-sdk-python/tree/main/doc) updated
- [X] __docstrings__: All public APIs documented

### Title and description

- [X] __Change type__: Title is prefixed with change type: and follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] __References__: Indicate issues fixed via: `Fixes #xxx`

----

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.

### DIFF
--- a/src/stepfunctions/steps/states.py
+++ b/src/stepfunctions/steps/states.py
@@ -254,25 +254,25 @@ class State(Block):
 
     def add_retry(self, retry):
         """
-        Add a Retry block to the tail end of the list of retriers for the state.
+        Add a Retry block or a list of Retry blocks to the tail end of the list of retriers for the state.
 
         Args:
-            retry (Retry): Retry block to add.
+            retry (Retry or list(Retry)): Retry block(s) to add.
         """
         if Field.Retry in self.allowed_fields():
-            self.retries.append(retry)
+            self.retries.extend(retry) if isinstance(retry, list) else self.retries.append(retry)
         else:
             raise ValueError("{state_type} state does not support retry field. ".format(state_type=type(self).__name__))
 
     def add_catch(self, catch):
         """
-        Add a Catch block to the tail end of the list of catchers for the state.
+        Add a Catch block or a list of Catch blocks to the tail end of the list of catchers for the state.
 
         Args:
-            catch (Catch): Catch block to add.
+            catch (Catch or list(Catch): Catch block(s) to add.
         """
         if Field.Catch in self.allowed_fields():
-            self.catches.append(catch)
+            self.catches.extend(catch) if isinstance(catch, list) else self.catches.append(catch)
         else:
             raise ValueError("{state_type} state does not support catch field. ".format(state_type=type(self).__name__))
 
@@ -487,10 +487,12 @@ class Parallel(State):
     A Parallel state causes the interpreter to execute each branch as concurrently as possible, and wait until each branch terminates (reaches a terminal state) before processing the next state in the Chain.
     """
 
-    def __init__(self, state_id, **kwargs):
+    def __init__(self, state_id, retry=None, catch=None, **kwargs):
         """
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
+            retry (Retry or list(Retry), optional): Retry block(s) to add to list of Retriers that define a retry policy in case the state encounters runtime errors
+            catch (Catch or list(Catch), optional): Catch block(s) to add to list of Catchers that define a fallback state that is executed if the state encounters runtime errors and its retry policy is exhausted or isn't defined
             comment (str, optional): Human-readable comment or description. (default: None)
             input_path (str, optional): Path applied to the state’s raw input to select some or all of it; that selection is used by the state. (default: '$')
             parameters (dict, optional): The value of this field becomes the effective input for the state.
@@ -499,6 +501,12 @@ class Parallel(State):
         """
         super(Parallel, self).__init__(state_id, 'Parallel', **kwargs)
         self.branches = []
+
+        if retry:
+            self.add_retry(retry)
+
+        if catch:
+            self.add_catch(catch)
 
     def allowed_fields(self):
         return [
@@ -536,11 +544,13 @@ class Map(State):
     A Map state can accept an input with a list of items, execute a state or chain for each item in the list, and return a list, with all corresponding results of each execution, as its output.
     """
 
-    def __init__(self, state_id, **kwargs):
+    def __init__(self, state_id, retry=None, catch=None, **kwargs):
         """
         Args:
             state_id (str): State name whose length **must be** less than or equal to 128 unicode characters. State names **must be** unique within the scope of the whole state machine.
             iterator (State or Chain): State or chain to execute for each of the items in `items_path`.
+            retry (Retry or list(Retry), optional): Retry block(s) to add to list of Retriers that define a retry policy in case the state encounters runtime errors
+            catch (Catch or list(Catch), optional): Catch block(s) to add to list of Catchers that define a fallback state that is executed if the state encounters runtime errors and its retry policy is exhausted or isn't defined
             items_path (str, optional): Path in the input for items to iterate over. (default: '$')
             max_concurrency (int, optional): Maximum number of iterations to have running at any given point in time. (default: 0)
             comment (str, optional): Human-readable comment or description. (default: None)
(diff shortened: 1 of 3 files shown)


src/stepfunctions/steps/states.py
```diff
@@ -254,27 +254,27 @@ def accept(self, visitor):
...
         Args:
-            retry (Retry): Retry block to add.
+            retry (Retry or list(Retry)): Retry block(s) to add.
         """
         if Field.Retry in self.allowed_fields():
-            self.retries.append(retry)
+            self.retries.extend(retry) if isinstance(retry, list) else self.retries.append(retry)
         else:
-            raise ValueError("{state_type} state does not support retry field. ".format(state_type=type(self).__name__))
+            raise ValueError(f"{type(self).__name__} state does not support retry field. ")
```
shivlaks, line 265: nice! love the use of `f strings to leave the code a little better than when you found it.

tests/unit/test_steps.py
```diff
@@ -469,4 +469,66 @@ def test_default_paths_not_converted_to_null():
...
+
+
+@pytest.mark.parametrize("state, state_id, extra_args, catch, expected_catch", [
+    (Parallel, 'Parallel', {}, CATCH, EXPECTED_CATCH),
+    (Parallel, 'Parallel', {}, CATCHES, EXPECTED_CATCHES),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, CATCH, EXPECTED_CATCH),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, CATCHES, EXPECTED_CATCHES),
+    (Task, 'Task', {}, CATCHES, EXPECTED_CATCHES),
+    (Task, 'Task', {}, CATCHES, EXPECTED_CATCHES)
+])
```
shivlaks, line 530: i must say, i'm only lukewarm on parameterized tests... as we get further and further away (in terms of locality) from the expectations and the properties, i find myself scrolling back and forth just to understand what we're even covering.

tests/unit/test_steps.py
```diff
@@ -469,4 +469,66 @@ def test_default_paths_not_converted_to_null():
...
+    (Parallel, 'Parallel', {}, RETRIES, EXPECTED_RETRIES),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, RETRIES, EXPECTED_RETRIES),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, RETRY, EXPECTED_RETRY),
+    (Task, 'Task', {}, RETRY, EXPECTED_RETRY),
+    (Task, 'Task', {}, RETRIES, EXPECTED_RETRIES)
+])
+def test_state_with_added_retry(state, state_id, extra_args, retry, expected_retry):
+    step = state(state_id, **extra_args)
+    step.add_retry(retry)
+    assert step.to_dict()['Retry'] == expected_retry
```
shivlaks, line 520: nit: can we keep retry tests together and the related constants a little more localized. It doesn't affect functionality in any way but significantly improves readability

tests/unit/test_steps.py
```diff
@@ -469,4 +469,66 @@ def test_default_paths_not_converted_to_null():
...
+
+@pytest.mark.parametrize("state, state_id, extra_args, retry, expected_retry", [
+    (Parallel, 'Parallel', {}, RETRY, EXPECTED_RETRY),
+    (Parallel, 'Parallel', {}, RETRIES, EXPECTED_RETRIES),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, RETRIES, EXPECTED_RETRIES),
+    (Map, 'Map', {'iterator': Pass('Iterator')}, RETRY, EXPECTED_RETRY),
+    (Task, 'Task', {}, RETRY, EXPECTED_RETRY),
+    (Task, 'Task', {}, RETRIES, EXPECTED_RETRIES)
+])
+def test_state_with_added_retry(state, state_id, extra_args, retry, expected_retry):
```
shivlaks, line 517: nit: not for this PR - I think we need to come up with a naming convention for our test suite. in this scenario, you're really just testing the "retry" part based on your assertions so why not simplify by calling it `test_add_retry` or something like that?

tests/integ/test_state_machine_definition.py
```diff
@@ -531,4 +584,56 @@ def test_retry_state_machine_creation(sfn_client, sfn_role_arn, training_job_par
...
+
+def test_state_machine_creation_with_retry_in_constructor(sfn_client, sfn_role_arn, training_job_parameters):
+    retry_state_name = "RetryStateName"
+    all_fail_error = "Starts.ALL"
+    interval_seconds = 1
+    max_attempts = 2
+    backoff_rate = 2
+    task_resource = f"arn:{get_aws_partition()}:states:::sagemaker:createTrainingJob.sync"
+
+    # change the parameters to cause task state to fail
```
shivlaks, line 598: specify what will cause it to fail...  i.e. "invalid image will throw badImageexception (or whatever)"

tests/integ/test_state_machine_definition.py
```diff
@@ -531,4 +584,56 @@ def test_retry_state_machine_creation(sfn_client, sfn_role_arn, training_job_par
         role=sfn_role_arn
     )
 
-    workflow_test_suite(sfn_client, workflow, asl_state_machine_definition, None)
\ No newline at end of file
+    workflow_test_suite(sfn_client, workflow, asl_state_machine_definition, None)
+
+
+def test_state_machine_creation_with_retry_in_constructor(sfn_client, sfn_role_arn, training_job_parameters):
```
shivlaks, line 590: nit: i don't think this constructor scenario needs it's own integ test, but we should have one for retry (or for catch). 
do we already have an existing test we could expand perhaps?

i'd suggest having a test that sets both a `retry` in the constructor as well as another through `add_retry` if possible. Having said that, if you can't think of a way to simplify, having a test is better than not having a test *always*

shivlaks (changes requested):
great job! 

couple of small general things
* this is a fix... the discarding of user provided parameters in constructor is a bug that we are addressing rather than a feature.
* let's tighten up the description to also include that this is for the `Task`, `Parallel`, and `Map` states. This is covered in the subsequent section but I think it belongs in the title
* "Manual Testing" - this should detail what manual testing we performed. Another contributor should be able to read your description and perform the same manual test. It's both a teaching tool as well as a valuable artifact for posterity.

PR #176, opened 2021-10-29
refactor: Update AWS Glue Databrew StartJobRun step to use integration pattern input

ca-nguyen:
### Description

Update AWS Glue Databrew service integration to use `integration_pattern` input instead or `wait_for_completion` flag.

Fixes #(issue) (N/A)

### Why is the change necessary?
This change is necessary for consistency with the new service integration implementation pattern introduced in [commit](https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/f9e63b312f543954c663e305fd5eedd9b2cdd360) (Add support for Nested Step Functions) that uses the `integration_pattern` arg in the step constructor to build the resource.

Support for AWS Glue Databrew service integration was added in this [commit](https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/ac652ab84ff74284ce996097c7c857e642043a3b), but not **released yet**.
A later [commit](https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/f9e63b312f543954c663e305fd5eedd9b2cdd360) (Add support for Nested Step Functions) introduced a new implementation pattern using the `IntegrationPattern` enum as input to construct the step instead of the `wait_for_completion` flag. (See [PR](https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/166) for more detail on rationale behind the implementation).

### Solution

 Replace the `wait_for_completion` flag with `integration_pattern` arg in [StartJobRun](https://docs.aws.amazon.com/step-functions/latest/dg/connect-databrew.html) step construction.
 
 The `IntegrationPattern` is used to build the `Resource` arn as follow:
| IntegrationPattern  | Resource                                                                                 | Doc
|:------------------------|:----------------------------------------------------------|:---|
| WaitForCompletion    | "arn:aws:states:::states:databrew:startJobRun.sync"                   | [Run A job](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html#connect-sync)|
| CallAndContinue                    | "arn:aws:states:::states:databrew:startJobRun"                                | [Request Response](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html#connect-default)|

See [Service Integration Patterns](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html) for more details
 
 Normally, replacing a constructor argument would be a breaking change, but since we have not released support for AWS Glue Databrew service integration yet, it is acceptable to do so. After next release, it making such changes will be considered as not being backward compatible.

### Testing

[...]

PR #180, opened 2021-11-12
Add ability to use waitForTaskToken for ECS tasks

nicolaei:
### Description

This change adds the ability to use `waitForTaskToken` for ECS tasks.

### Why is the change necessary?

To add support for documentet features. Ref [the ECS step docs in the step function documentation](https://docs.aws.amazon.com/step-functions/latest/dg/connect-ecs.html).

### Solution

The solution uses the already established pattern of having an argument `wait_for_callback`, as seen in the lambda task.

But because we can't have both `wait_for_completion` and `wait_for_callback` at the same time, I've added a condition to raise a `ValueError` if both are set to `True`. I'm happy to change this if you've already got an established pattern for argument errors.

### Testing

Created a Unittest and also using this in a project at work.

----

### Pull Request Checklist

Please check all boxes (including N/A items)

#### Testing

- [x] Unit tests added
- [x ] Integration test added
- [x] Manual testing - why was it necessary? could it be automated?

#### Documentation

- [x] __docs__: All relevant [docs](https://github.com/aws/aws-step-functions-data-science-sdk-python/tree/main/doc) updated
- [x] __docstrings__: All public APIs documented

### Title and description

- [x] __Change type__: Title is prefixed with change type: and follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] __References__: Indicate issues fixed via: `Fixes #xxx`

----

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache-2.0 license.


wong-a:
Thanks for the PR! For integrations that support all 3 integration patterns, we started to introduce a different way to set this using the [`IntegrationPattern`](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/integration_resources.py#L20-L27) enum instead of separate args for each pattern.

- https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/f9e63b312f543954c663e305fd5eedd9b2cdd360
- https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/176

nicolaei:
Sorry about the multiples of force-pushes (and subsequent spams from the build). Had some issues with the test-suite on my local machine for some reason 🤔

> Thanks for the PR! For integrations that support all 3 integration patterns, we started to introduce a different way to set this using the [`IntegrationPattern`](https://github.com/aws/aws-step-functions-data-science-sdk-python/blob/main/src/stepfunctions/steps/integration_resources.py#L20-L27) enum instead of separate args for each pattern.
> 
> * [f9e63b3](https://github.com/aws/aws-step-functions-data-science-sdk-python/commit/f9e63b312f543954c663e305fd5eedd9b2cdd360)
> * [refactor: Update AWS Glue Databrew StartJobRun step to use integration pattern input #176](https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/176)

Would you like me to change the PR to use that pattern? I could introduce it to all the other task classes in `compute.py` while I'm at it if you'd like.

wong-a (changes requested):

wong-a:
> Would you like me to change the PR to use that pattern?
> Seeing that it would be a breaking change as well, I'd probably have to change the PR title to breaking?

See  https://github.com/aws/aws-step-functions-data-science-sdk-python/pull/180#discussion_r748630286

>  I could introduce it to all the other task classes in compute.py while I'm at it if you'd like.
We like to keep PRs small, so just ECS is more than fine. The main motivation here is supporting all 3 integration patterns and ECS is the only implemented integration that doesn't expose all of them.

nicolaei:
Sorry for the late change here, but here you go!

nicolaei:
@wong-a: Any chance we can get this merged? 😄

jormello (changes requested):
Thanks for addressing the feedback from @wong-a and for reaching back out for an update @nicolaei. 

Source code changes look okay to me. For the test code changes, can you add test cases for each of the integration patterns? Right now there's only coverage for `IntegrationPattern.WaitForTaskToken`.

nicolaei:
I updated the tests with explicit testing of both old-style and new-style parameters now!

I'm a bit confused why the tests are failing though, because these build-issues seem totally unrelated, and I only updated my test-code 🤔 Any thoughts @jormello?

